PR #1, opened 2019-08-31
base: Use Fedora 30 for godot-fedora:latest

akien-mga:
Note: I branched off the current `master` branch to 3.1, so that we can keep the same containers for 3.1.x and upgrade them for 3.2/master.


hpvb:
Thanks!

PR #2, opened 2019-08-31
javascript: Cleanup and source env in ~/.bashrc

akien-mga:
This way we shouldn't need to source `emsdk_env.sh` manually when building templates.


akien-mga:
This one should be OK to merge @hpvb.

hpvb:
Thanks!

PR #3, opened 2019-08-31
ubuntu: Use xz-utils for mono 6.0.x

akien-mga:
Since 6.0.0.x upstream changed from `tar.bz2` to `tar.xz`: https://download.mono-project.com/sources/mono/

I think we should use the latest stable for 3.2, so this change makes sense in the master branch.

Note that Ubuntu 14.04.x is EOL, and Mono upstream stopped providing new packages for msbuild since one year. Right now it seems to still be working, but I guess eventually @neikeq might need a more recent msbuild.

I've started work on a CentOS 7 based container, which is still supported by RedHat, and for which Mono upstream still provides new packages. With the [SCL](https://wiki.centos.org/AdditionalResources/Repositories/SCL) we can have GCC 8, which is nice, and we get an even older glibc than on Ubuntu, so also good for back compat. One drawback: CentOS doesn't provide the SCL for i686, so the options would be:
- keep `ubuntu-32` for this Linux 32-bit
- drop 32-bit Linux support from official builds
- build 32-bit Linux binaries against CentOS 7's older toolchain

I'll make a PR with the 64-bit CentOS 7 dockerfile.

### DIFF
--- a/Dockerfile.ubuntu-32
+++ b/Dockerfile.ubuntu-32
@@ -13,12 +13,12 @@ RUN apt-get update && \
     echo 'deb http://ppa.launchpad.net/mc3man/trusty-media/ubuntu trusty main' >> /etc/apt/sources.list && \
     apt-get update && \
     apt-get install -y gcc-8 g++-8 libudev-dev libx11-dev libxcursor-dev libxrandr-dev libasound2-dev libpulse-dev \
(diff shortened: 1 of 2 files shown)


akien-mga:
Ah and `MONO32_PREFIX` and `MONO64_PREFIX` are removed from the env as they're no longer used for 3.2/master. (And were not needed anyway since we can locate Mono via pkgconfig as it's installed in `/usr`.)

akien-mga:

PR #4, opened 2019-08-31
[WIP] linux: Add CentOS 7 based container for x86_64

akien-mga:
Thanks to the SCL, we can get the GCC 8 toolchain,
and GCC 9 should be made available once 9.3 is released.

The setup is cleaner than on Ubuntu and we have an even
older glibc version for compatibility. The drawback is
that there is no SCL for CentOS 7 i686, and Mono upstream
also doesn't provide packages for it (though we could build
them from the src.rpm).

And of course, Ubuntu 14.04 is now EOL so it makes sense to
make our builds on a still supported release.

---

As mentioned on #3, if we want to use CentOS 7 for x86_64 Linux builds, we might still need to keep `ubuntu-32` for i686 builds, as CentOS doesn't seem to provide the GCC 8 toolchain out of the box for it.

I did find the `devtoolset-7` for RHEL7 i686 on the CloudLinux repo though, so I guess it might still possible to get this somehow? https://www.repo.cloudlinux.com/cloudlinux/7.6/sclo/devtoolset-7/i386/

### DIFF
--- a/Dockerfile.linux-64
+++ b/Dockerfile.linux-64
@@ -1,0 +1,32 @@
+FROM centos:7
+
+ARG mono_version
+
+RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
+    yum update && \
+    yum install -y centos-release-scl && \
+    yum install -y devtoolset-8-toolchain && \
+    echo "source scl_source enable devtoolset-8" >> /root/.bashrc && \
+    yum install -y git scons pkgconfig libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel libXi-devel \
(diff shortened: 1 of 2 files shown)


akien-mga:
Updated to add a missing `cd /root`, due to which the mono build dir didn't get properly removed (as I assumed it was in `/root/mono-${mono_version}` when it was in `/mono-${mono_version}`.

akien-mga:
Marking this one WIP for now as we don't have a good solution for i686 support yet.

Conan-Kudo:
@akien-mga There is an i686 altarch build for CentOS 7, but I don't know if devtoolset was built for it.

akien-mga:
Superseded by #80, now using Fedora images with a `buildroot` toolchain for all supported architectures: https://github.com/godotengine/buildroot/

PR #5, opened 2019-08-31
mono: Add support for mono 6.0.x

akien-mga:
This should actually be merged before #3 and #4 which build upon this image.

Mono 6.0.x now ships as `.tar.xz` so some changes were needed. It also seems to have a bogus `configure` by default, so we fix it with `autoreconf -vfi`.

I also bumped all msbuild packages to their latest version from `centos8-stable`.

### DIFF
--- a/Dockerfile.mono
+++ b/Dockerfile.mono
@@ -4,15 +4,21 @@ ARG mono_version
 
 RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi
 
-RUN dnf -y install cmake gcc gcc-c++ make which perl python curl bzip2 && dnf clean all && \
-    curl https://download.mono-project.com/sources/mono/mono-${mono_version}.tar.bz2 | tar xj && \
+RUN dnf -y install autoconf automake libtool cmake gcc gcc-c++ make which perl python curl bzip2 xz && dnf clean all && \
+    if [ ${mono_version%%.*} -ge 6 ]; then \
+        curl https://download.mono-project.com/sources/mono/mono-${mono_version}.tar.xz | tar xJ; \
+    else \
+        curl https://download.mono-project.com/sources/mono/mono-${mono_version}.tar.bz2 | tar xj; \
+    fi && \
     cd mono-${mono_version} && \
-    ./configure --prefix=/usr --sysconfdir=/etc --localstatedir=/var/lib/mono --disable-boehm && make -j && make install && make distclean && \
+    autoreconf -vfi && \
+    ./configure --prefix=/usr --sysconfdir=/etc --localstatedir=/var/lib/mono --disable-boehm && \
+    make -j && make install && make distclean && \
     cd /root && \
     rm -rf mono-${mono_version} && \
-    cert-sync /etc/pki/tls/certs/ca-bundle.crt && \
-    rpm -ivh --nodeps https://download.mono-project.com/repo/centos7-stable/m/msbuild/msbuild-16.0+xamarinxplat.2018.09.26.17.53-0.xamarin.3.epel7.noarch.rpm \
-                      https://download.mono-project.com/repo/centos7-stable/m/msbuild-libhostfxr/msbuild-libhostfxr-2.0.0.2017.07.06.00.01-0.xamarin.1.epel7.x86_64.rpm \
-                      https://download.mono-project.com/repo/centos7-stable/m/msbuild/msbuild-sdkresolver-16.0+xamarinxplat.2018.09.26.17.53-0.xamarin.3.epel7.noarch.rpm \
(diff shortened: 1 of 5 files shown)


akien-mga:
> The hotfixed mingw-binutils packages should likely be rebuilt against
Fedora 30's mingw-binutils, but I didn't find the src.rpm in the repo
to do it. There were no functional changes between F29 and F30's
mingw-binutils though, so we should be fine as is:
https://src.fedoraproject.org/rpms/mingw-binutils/commits/f30

I just noticed that the `src.rpm` file *is* in the repo, so I'll work on rebuilding the packages properly for F30.

akien-mga:
The msbuild update seems to cause issue when building mono 5.18.1.3 (same as for 3.1.x builds), as they changed the packaging from `/usr/lib/mono/msbuild/15.0` to `/usr/lib/mono/msbuild/Current`, with a symlink from `15.0` to `Current`.

That's an issue because the source build of 5.18.1.3 installs Roslyn to `/usr/lib/mono/msbuild/15.0/bin/Roslyn/`, and RPM rightfully fails when it tries to set up the `15.0` symlink but it's already a real folder. As such it moves `15.0/bin/Roslyn` to `15.0.rpmmoved/bin/Roslyn`, which breaks msbuild.

So this PR should be merged if we want to build a more recent Mono, but not for 5.18.1.3.

akien-mga:
> I also bumped all msbuild packages to their latest version from `centos8-stable`.

I reverted this change following https://github.com/godotengine/build-containers/pull/5#issuecomment-527097913

We can redo this update once we decide to update all build containers to mono 6.0.x, provided we also switch the Linux build containers from Ubuntu 14.04 to CentOS 7.

Dockerfile.mono
```diff
@@ -4,15 +4,21 @@ ARG mono_version
...
     cd /root && \
     rm -rf mono-${mono_version} && \
-    cert-sync /etc/pki/tls/certs/ca-bundle.crt && \
-    rpm -ivh --nodeps https://download.mono-project.com/repo/centos7-stable/m/msbuild/msbuild-16.0+xamarinxplat.2018.09.26.17.53-0.xamarin.3.epel7.noarch.rpm \
-                      https://download.mono-project.com/repo/centos7-stable/m/msbuild-libhostfxr/msbuild-libhostfxr-2.0.0.2017.07.06.00.01-0.xamarin.1.epel7.x86_64.rpm \
-                      https://download.mono-project.com/repo/centos7-stable/m/msbuild/msbuild-sdkresolver-16.0+xamarinxplat.2018.09.26.17.53-0.xamarin.3.epel7.noarch.rpm \
-                      https://download.mono-project.com/repo/centos7-stable/n/nuget/nuget-4.7.0.5148.bin-0.xamarin.1.epel7.noarch.rpm
+    cert-sync /etc/pki/tls/certs/ca-bundle.crt
 
+RUN rpm -ivh --nodeps https://download.mono-project.com/repo/centos8-stable/m/msbuild/msbuild-16.0+xamarinxplat.2018.09.26.17.53-0.xamarin.5.epel8.noarch.rpm \
```
akien-mga: FYI, I made this a separate `RUN` step as I had some issues with folder conflicts depending on the mono and RPM versions, so to debug this it's simpler to have an intermediate image before that.

Dockerfile.mono
```diff
@@ -4,15 +4,21 @@ ARG mono_version
 
 RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi
 
-RUN dnf -y install cmake gcc gcc-c++ make which perl python curl bzip2 && dnf clean all && \
-    curl https://download.mono-project.com/sources/mono/mono-${mono_version}.tar.bz2 | tar xj && \
+RUN dnf -y install autoconf automake libtool cmake gcc gcc-c++ make which perl python curl bzip2 xz && dnf clean all && \
+    if [ ${mono_version%%.*} -ge 6 ]; then \
```
akien-mga: As mentioned on https://github.com/godotengine/build-containers/pull/7#issuecomment-536419525, I had to use a Git clone for the Android build, so I might as well do the same for all the builds here?
That also removes the awkwardness of handling the switch from bzip to xz in 6.0 :P

Dockerfile.windows
```diff
@@ -4,35 +4,43 @@ ARG mono_version
...
+    else \
+        curl https://download.mono-project.com/sources/mono/mono-${mono_version}.tar.bz2 | tar xj; \
+    fi && \
+    cp -r mono-${mono_version} mono-${mono_version}-32 && \
     cd mono-${mono_version} && \
     ./configure --prefix=/root/dependencies/mono-64 --host=x86_64-w64-mingw32 --disable-boehm --disable-mcs-build --disable-executables && \
     echo '#define HAVE_STRUCT_SOCKADDR_IN6 1' >> config.h && \
     make -j && \
     make install && \
-    make distclean && \
```
akien-mga: I had to change this to work around mono/mono#16605.

PR #6, opened 2019-08-31
windows: Use xz for mono 6.0.x, workaround distclean bug

akien-mga:
Same as #5, since the mono 6.0+ tarballs use .tar.xz, we need
some changes.

I also run into a build issue due to `make distclean` being bogus
in 6.0.0.319 (mono/mono#16605), so I worked it around by starting
from a clean folder for each build.

The hotfixed binutils packages should likely be rebuilt against
Fedora 30's mingw-binutils, but I didn't find the src.rpm in the
repo to do it. There were no functional changes between F29 and
F30's mingw-binutils though, so we should be fine as is:
https://src.fedoraproject.org/rpms/mingw-binutils/commits/f30

The `MONO32_PREFIX` and `MONO64_PREFIX` are no longer used by
the buildsystem (replaced by a `mono_prefix` build option), but
I kept them as they're convenient to point to the relevant install
locations.

### DIFF
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -4,35 +4,41 @@ ARG mono_version
 
 RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi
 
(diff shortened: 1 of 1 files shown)


akien-mga:

PR #7, opened 2019-08-31
android: Set up Mono buildsystem

akien-mga:
Adds support for building Mono templates for Android in 3.2/master.
Expects mono 6.0+ (otherwise we'd have to handle the tarball compression
change from upstream).

We use the script from https://github.com/godotengine/godot-mono-builds
to build Mono for all arches with the right options, and a patch included
in https://github.com/godotengine/godot/blob/master/modules/mono/build_scripts/patches/fix-mono-android-tkill.diff
which is also necessary.

I also moved the SDK to a `sdk` folder to make the `$HOME` more tidy,
especially since the Mono build scripts install the mono toolchains in
`$HOME` too by default.

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -1,13 +1,30 @@
-FROM godot-fedora:latest
+ARG mono_version
+FROM godot-mono:${mono_version}
+ARG mono_version
 
 RUN dnf -y install scons java-1.8.0-openjdk-devel ncurses-compat-libs unzip which gcc gcc-c++ && \
+    mkdir sdk && cd sdk && \
     curl -LO https://dl.google.com/android/repository/sdk-tools-linux-4333796.zip && \
     unzip sdk-tools-linux-4333796.zip && \
     rm sdk-tools-linux-4333796.zip && \
     yes | tools/bin/sdkmanager --licenses && \
-    tools/bin/sdkmanager ndk-bundle 'platforms;android-23' 'build-tools;19.1.0' 'build-tools;28.0.3' 'platforms;android-28' 
+    tools/bin/sdkmanager ndk-bundle 'build-tools;28.0.3' 'platforms;android-28' 'cmake;3.10.2.4988404' && \
+    cd ..
 
-ENV ANDROID_HOME=/root/
-ENV ANDROID_NDK_ROOT=/root/ndk-bundle/
+ENV ANDROID_HOME=/root/sdk/
+ENV ANDROID_NDK_ROOT=/root/sdk/ndk-bundle/
+
+RUN dnf -y install git patch && \
+    git clone https://github.com/mono/mono --branch mono-${mono_version} --single-branch && \
+    cd mono && git submodule update --init && \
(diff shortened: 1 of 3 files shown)


Dockerfile.android
```diff
@@ -1,13 +1,35 @@
...
+
+RUN dnf -y install git patch && \
+    if [ ${mono_version%%.*} -ge 6 ]; then \
+        curl https://download.mono-project.com/sources/mono/mono-${mono_version}.tar.xz | tar xJ; \
+    else \
+        curl https://download.mono-project.com/sources/mono/mono-${mono_version}.tar.bz2 | tar xj; \
+    fi && \
+    cd mono-${mono_version} && \
+    mono_sources=$(pwd) && \
+    curl -LO https://raw.githubusercontent.com/godotengine/godot/master/modules/mono/build_scripts/patches/fix-mono-android-tkill.diff && \
```
hpvb: Can you place this patch in files/ instead?

Dockerfile.android
```diff
@@ -1,13 +1,35 @@
...
+    patch -p1 < fix-mono-android-tkill.diff && \
+    cd .. && \
+    git clone https://github.com/godotengine/godot-mono-builds && \
```
hpvb: Can you check out a particular revision here? The containers need to be rebuildable by users.

akien-mga: Done.

Dockerfile.android
akien-mga: Done.

akien-mga:
I did the requested changes.

I also added some changes I had locally but not pushed to git yet, namely changing from using the upstream tarball to using a git clone for a specific release tag.
Mono upstream seems to have bugs in their tarball generation, and the 5.18.1.3 tarball lacked two files needed to build on Android, even though they are included in the 5.18.1.3 tag:
```
/root/mono-5.18.1.3/mono/metadata/../../support/libm/complex.c:19:10: fatal error: 'math_private.h' file not found
#include "math_private.h"
         ^~~~~~~~~~~~~~~~
```

So far this was only done for Android builds, but we could also use the same approach for `Dockerfile.mono` if you agree.

Dockerfile.android
```diff
@@ -1,13 +1,30 @@
...
+    tools/bin/sdkmanager ndk-bundle 'build-tools;28.0.3' 'platforms;android-28' 'cmake;3.10.2.4988404' && \
+    cd ..
 
-ENV ANDROID_HOME=/root/
-ENV ANDROID_NDK_ROOT=/root/ndk-bundle/
+ENV ANDROID_HOME=/root/sdk/
+ENV ANDROID_NDK_ROOT=/root/sdk/ndk-bundle/
+
+RUN dnf -y install git patch && \
+    git clone https://github.com/mono/mono --branch mono-${mono_version} --single-branch && \
```
akien-mga: For the reference, while we use `--branch` here, we check out a git tag, so unless upstream deletes it and retags another commit, this should allow reproducible builds.
The `git submodule update --init` below should also update to the commit hashes referenced in the tagged mono commit, so they shouldn't change.

PR #8, opened 2019-08-31
osx: Use xz for mono 6.0.x

akien-mga:
That's the only updated container which I couldn't test building so far, as I don't have access to the relevant SDK.

We'll need to increase our target SDK anyway for 3.2 to keep up with Apple's depreciation of older versions, and iOS will also need a version bump to support ARKit.


akien-mga:

Issue #13, opened 2019-11-13
Use consistent python version to improve rebuild speeds


tinmanjuggernaut:
Can we move towards using a consistent python major.minor version across all dockers? 

I build from the same godot source tree mounted in each. Rebuilding a godot platform is greatly sped up by scons tracking what has already been built in `.sconsign.dblite`. 

However, some dockers leave that file in a state that other dockers cannot read. The build dies with `scons: *** [AUTHORS.md] ValueError : unsupported pickle protocol: 4`. Pickle protocol 4 was released in python 3.5. Deleting the db fixes the error but adds on 20+ minutes of build time for rebuilds when it could take 1-2.

akien-mga:
Fixed by godotengine/godot@7f62bed2387db01e825b8d61253619e948c65058.

PR #14, opened 2019-11-13
Fix Docker.javascript not building

tinmanjuggernaut:
Docker.javascript fails to build

```
Cloning into 'emsdk'...
remote: Enumerating objects: 25, done.        
remote: Counting objects: 100% (25/25), done.        
remote: Compressing objects: 100% (21/21), done.        
remote: Total 1694 (delta 11), reused 7 (delta 4), pack-reused 1669        
Receiving objects: 100% (1694/1694), 960.21 KiB | 492.00 KiB/s, done.
Resolving deltas: 100% (1068/1068), done.
./emsdk: line 18: exec: python: not found
```

This docker installs scons which pulls in python3. There is no `python` command.

How do you want to handle it? Especially regarding #13. 
* Add `python` to the list of requested packages (which pulls in python2 [2.7.17])?
* `ln -s python3 python`?

Also, due to this emsdk bug, 
https://github.com/godotengine/godot/issues/33374 
shouldn't we specify a version like 1.39.0?

### DIFF
--- a/Dockerfile.javascript
+++ b/Dockerfile.javascript
@@ -1,10 +1,10 @@
 FROM godot-fedora:latest
 
-RUN dnf -y install scons git bzip2 xz java-openjdk yasm && dnf clean all && \
+RUN dnf -y install which scons git bzip2 xz java-openjdk yasm && dnf clean all && \
     git clone https://github.com/emscripten-core/emsdk && \
     cd emsdk && \
-    ./emsdk install latest && \
-    ./emsdk activate latest && \
+    ./emsdk install 1.39.0 && \
+    ./emsdk activate 1.39.0 && \
     echo "source /root/emsdk/emsdk_env.sh" >> /root/.bashrc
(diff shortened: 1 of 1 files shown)


akien-mga:
That's weird, it build fine as is last time I used it (a couple of months ago). I'll look into it; I assume that emsdk should be able to use python3 just fine.

Specifying a version makes sense, especially since we want to have reproducible builds.

akien-mga:
emsdk indeed has code to handle this properly: https://github.com/emscripten-core/emsdk/blob/master/emsdk#L11-L16

What we lack is not python2, but the `which` command.

tinmanjuggernaut:
Updated and tested a rebuild.

akien-mga:
Thanks!

hpvb:
`which` should probably just be added to godot-fedora instead, it's needed for mono also iirc.

Issue #15, opened 2019-11-13
Docker.osx fails to build


tinmanjuggernaut:
I put `Xcode-10.3.xip` in files/ then let it build godot-xcode-packer. Now my files/ also contains `iPhoneOS12.4.sdk.tar.xz`, `iPhoneSimulator12.4.sdk.tar.xz`, and `MacOSX10.14.sdk.tar.xz`.

Godot-osx fails here:
```
 /usr/bin/mkdir -p '/root/osxcross/target/share/man/man5'
 /usr/bin/install -c -m 644 'Mach-O.5' '/root/osxcross/target/share/man/man5/x86_64-apple-darwin17-Mach-O.5'
 /usr/bin/install -c -m 644 'a.out.5' '/root/osxcross/target/share/man/man5/x86_64-apple-darwin17-a.out.5'
 /usr/bin/install -c -m 644 'ranlib.5' '/root/osxcross/target/share/man/man5/x86_64-apple-darwin17-ranlib.5'
 /usr/bin/install -c -m 644 'stab.5' '/root/osxcross/target/share/man/man5/x86_64-apple-darwin17-stab.5'
make[2]: Leaving directory '/root/cctools-port/cctools/man'
make[1]: Leaving directory '/root/cctools-port/cctools/man'
make[1]: Entering directory '/root/cctools-port/cctools'
make[2]: Entering directory '/root/cctools-port/cctools'
make[2]: Nothing to be done for 'install-exec-am'.
make[2]: Nothing to be done for 'install-data-am'.
make[2]: Leaving directory '/root/cctools-port/cctools'
make[1]: Leaving directory '/root/cctools-port/cctools'
you must run ./build.sh first before you can start building compiler-rt
[...]

PR #16, opened 2019-11-15
Install common dependencies in Dockerfile.base

akien-mga:
Since we need to install those deps in most leaf containers, I figure it might be better to simply get them in the base one.

We can discuss whether all those should be included in the base or if some should be left in leaf containers, as well as whether we want to include more stuff in the base container (perl? nano? :P).

We also have a couple containers installing `python`, but I'm not sure it's really needed. We should likely aim for being python3-only (see also #13).

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -2,7 +2,9 @@ ARG mono_version
 FROM godot-mono:${mono_version}
 ARG mono_version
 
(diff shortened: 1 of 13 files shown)


akien-mga:
Tested a few times rebuilding all containers (apart from OSX due to #15, and I didn't try UWP), everything works as expected.

PR #17, opened 2019-11-15
Use git tag to checkout stable mono releases

akien-mga:
The main reason is that upstream Mono tends to miss some
important files in its tarballs, so Git tags appears to be
a more reliable medium.

Also split RUN further to do cleanup in a separate stage,
allowing to inspect intermediary images to debug build or
packaging issues.

---

**Depends on #16.**

Includes commits from #16, I'll rebase once that first PR is merged.

Only the last commit in this PR is meant to be reviewed here.

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -2,7 +2,8 @@ ARG mono_version
 FROM godot-mono:${mono_version}
 ARG mono_version
 
-RUN dnf -y install --setopt=install_weak_deps=False \
+RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
+    dnf -y install --setopt=install_weak_deps=False \
       gcc gcc-c++ java-1.8.0-openjdk-devel ncurses-compat-libs && \
     dnf clean all && \
     mkdir sdk && cd sdk && \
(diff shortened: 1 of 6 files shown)


hpvb:
Splitting RUN steps massively increases the size of images (at least when built with docker). Each command in a dockerfile creates a layer. So if you create a file in one command and delete it in the next you're still downloading that file.

akien-mga:
> Splitting RUN steps massively increases the size of images (at least when built with docker). Each command in a dockerfile creates a layer. So if you create a file in one command and delete it in the next you're still downloading that file.

Yeah, it's just convenient for debugging to be able to inspect intermediate stages. But we'd need to have a reliable way to delete/merge intermediate layers eventually to reduce disk usage. If that's not possible, then yeah, merging all RUNs together probably is the best option, with manually splitting when debugging.

hpvb:
> > Splitting RUN steps massively increases the size of images (at least when built with docker). Each command in a dockerfile creates a layer. So if you create a file in one command and delete it in the next you're still downloading that file.
> 
> Yeah, it's just convenient for debugging to be able to inspect intermediate stages. But we'd need to have a reliable way to delete/merge intermediate layers eventually to reduce disk usage. If that's not possible, then yeah, merging all RUNs together probably is the best option, with manually splitting when debugging.

We can post-process them, or build the production images with podman. Even then though, we need to have multiple *images* per step otherwise there will be no layers at all. This would mean that the base OS will get downloaded in its entirety with each container.

Dockerfiles are really, really dumb. We could just drop support for building with Docker entirely and use buildah directly. This would let us do pretty much whatever we want, and the images would still run with docker. But this would make it very difficult for Windows users to rebuild the private containers.

akien-mga:
Well I pushed an update to undo my adding of extra RUN steps (and I grouped a few more together to reduce the number of layers). It's easy enough to add extra RUNs manually when I need to log into an intermediate layer to debug things like file conflicts when installing the msbuild RPMs/DEBs.

akien-mga:
Tested successfully on all affected container builds (apart from OSX due to build issue, though I tested the changes indirectly in the existing OSX container).

tinmanjuggernaut:
> Dockerfiles are really, really dumb. We could just drop support for building with Docker entirely and use buildah directly. This would let us do pretty much whatever we want, and the images would still run with docker. But this would make it very difficult for Windows users to rebuild the private containers.

I'm new to the docker world, but not to vm. What if we released pre-built docker images? Isn't that a thing in docker-land? Then one could just download the images they need to compile against and save 10 hours of compiling, and avoid broken builds when some dependent repository got updated.

PR #18, opened 2019-11-15
Build Mono WebAssembly runtime and BCLs

akien-mga:
Up until now we used the desktop BCL (compiled with desktop Mono)
for both desktop and Android, but that does not work for WebAssembly
(and we suspect Android issues related to it).

So we now build the desktop, Android and WebAssembly BCLs in the
relevant containers, which are later copied in the templates directory
to include on export.

We don't need to build the desktop BCL per se as we already have it
installed somewhere in the Mono root, but I figured it will be easier to
maintain for us to have things done the same way for all platforms.

---

**Depends on #16 and #17.**

I'll rebase once those are merged. I just want all in the same local branch to be able to actually test my changes with a full rebuild.


akien-mga:
Pushed a few iterations after testing and debugging most of yesterday. We found with @neikeq that Emscripten 1.38.48+ breaks the mono support for us (`.wasm` file doesn't get generated), so I'm now pinning the version to 1.38.47 (which works).

akien-mga:
Tested successfully on all affected container builds (apart from OSX due to build issue, though I tested the changes indirectly in the existing OSX container).

PR #19, opened 2019-11-18
Use /bin/bash as default CMD, log container build output

akien-mga:
The previous `['/bin/bash']` was not working, running the image
with `podman run -it <name>` would evaluate to:
```
> sh -c ['/bin/bash']
sh: [/bin/bash]: No such file or directory
```


akien-mga:
Tested successfully over the past few days.

PR #23, opened 2019-11-21
Refactor image versioning, using $godot_branch-$mono_version format

akien-mga:
E.g. `3.1-mono-5.18.1.3` or `master-2019-08`.
The images which do not include mono only use the Godot branch,
i.e. `3.1` or `master`.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,29 +13,31 @@ if ! which $podman; then
 fi
 
 if [ -z "$1" ]; then
-  echo "usage: $0 <mono version"
+  echo "usage: $0 <godot branch> <mono git version>"
   echo
-  echo "For example: $0 5.16.0.220"
+  echo "For example: $0 3.1 mono-5.18.1.3"
   echo
   exit 1
 fi
 
-mono_version=$1
+godot_branch=$1
+mono_version=$2
+img_version=$godot_branch-$mono_version
 
 mkdir -p logs
 
-$podman build -t godot-fedora:latest -f Dockerfile.base . 2>&1 | tee logs/base.log
-$podman build -t godot-export:latest -f Dockerfile.export . 2>&1 | tee logs/export.log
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -13,29 +13,31 @@ if ! which $podman; then
...
+$podman build --build-arg mono_version=${mono_version} -t godot-mono:${img_version} -f Dockerfile.mono . 2>&1 | tee logs/mono.log
+$podman build --build-arg mono_version=${mono_version} -t godot-mono-glue:${img_version} -f Dockerfile.mono-glue . 2>&1 | tee logs/mono-glue.log
+$podman build --build-arg mono_version=${mono_version} -v $(pwd)/files:/root/files -t godot-windows:${img_version} -f Dockerfile.windows . 2>&1 | tee logs/windows.log
+$podman build --build-arg mono_version=${mono_version} -t godot-ubuntu-64:${img_version} -f Dockerfile.ubuntu-64 . 2>&1 | tee logs/ubuntu-64.log
+$podman build --build-arg mono_version=${mono_version} -t godot-ubuntu-32:${img_version} -f Dockerfile.ubuntu-32 . 2>&1 | tee logs/ubuntu-32.log
+$podman build --build-arg mono_version=${mono_version} -t godot-android:${img_version} -f Dockerfile.android . 2>&1 | tee logs/android.log
+$podman build --build-arg mono_version=${mono_version} -v $(pwd)/files:/root/files -t godot-javascript:${img_version}-upstream -f Dockerfile.javascript . 2>&1 | tee logs/javascript.log
```
akien-mga: Ah I added `-upstream` here as emscripten has two backends, `fastcomp` (old one) and `upstream` (llvm upstream, new default since 1.39.0). Since we're packaging a version just before the official switch to `upstream` I put this suffix to make things explicit. Eventually we might not want to bother having it since `upstream` will become the norm.

PR #25, opened 2019-11-23
OSX: Fix config file for mono 6.0+

akien-mga:
This effectively means that we can't use the current Dockerfile.osx
to build mono < 6.0, as the DLL path changed.

I've asked users on Discord to confirm that this fixes the issue, will merge if they do.

Related issue (for Windows): godotengine/godot#33822.

### DIFF
--- a/Dockerfile.osx
+++ b/Dockerfile.osx
@@ -38,8 +38,10 @@ RUN git clone https://github.com/mono/mono --branch ${mono_version} --single-bra
     ln -s /usr/bin/mono /root/dependencies/mono/bin/mono && \
     ln -s /usr/bin/mono-sgen /root/dependencies/mono/bin/mono-sgen && \
     ln -sf /usr/lib/mono/* /root/dependencies/mono/lib/mono && \
-    cp -rvp /etc/mono /root/dependencies/mono/etc && \
-    cp /root/files/mono-config-macosx /root/dependencies/mono/etc/config && \
+    mkdir -p /root/dependencies/mono/etc && \
+    cp -rvp /etc/mono /root/dependencies/mono/etc/ && \
+    cp /root/dependencies/mono/etc/mono/config{,.bak} && \
+    cp /root/files/mono-config-macosx /root/dependencies/mono/etc/mono/config && \
     rm -rf /root/mono
 
 ENV MONO64_PREFIX=/root/dependencies/mono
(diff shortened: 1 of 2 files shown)


akien-mga:
BTW, we still have a few `.so` libs referenced in this file, notably GTK ones, I don't know if it's fine or if they should be changed. For this PR I only did the same changes as done in the Linux/Windows configs.

Dockerfile.osx
```diff
@@ -38,8 +38,10 @@ RUN git clone https://github.com/mono/mono --branch ${mono_version} --single-bra
     ln -s /usr/bin/mono /root/dependencies/mono/bin/mono && \
     ln -s /usr/bin/mono-sgen /root/dependencies/mono/bin/mono-sgen && \
     ln -sf /usr/lib/mono/* /root/dependencies/mono/lib/mono && \
-    cp -rvp /etc/mono /root/dependencies/mono/etc && \
```
akien-mga: We were again copying the structure wrongly if `/root/dependencies/mono/etc` didn't exist, as the copied `mono` folder would become the `${prefix}/etc` folder instead of `${prefix}/etc/mono`.

Dockerfile.osx
```diff
@@ -38,8 +38,10 @@ RUN git clone https://github.com/mono/mono --branch ${mono_version} --single-bra
...
-    cp /root/files/mono-config-macosx /root/dependencies/mono/etc/config && \
```
akien-mga: We copied this file to the wrong location, it's meant to be in `${prefix}/etc/mono/config`.
The in-container build script then used the wrong path too so it worked, but it's better to use the expected path in the mono distribution to avoid confusion (and I back up the old file for comparison, as new Mono versions sometimes add content there).

hpvb:
> BTW, we still have a few `.so` libs referenced in this file, notably GTK ones, I don't know if it's fine or if they should be changed. For this PR I only did the same changes as done in the Linux/Windows configs.

That file came from a normal OSX mono installation so it shouldn't be an issue.

hpvb:
> This effectively means that we can't use the current Dockerfile.osx
> to build mono < 6.0, as the DLL path changed.
> 
> I've asked users on Discord to confirm that this fixes the issue, will merge if they do.
> 
> Related issue (for Windows): [godotengine/godot#33822](https://github.com/godotengine/godot/issues/33822).

When we bundle mono (which we do for our builds) we get to decide where these files are due to this config file. So we could massage either one to have the same name as the other.
Alternatively since we have a variable input for the mono version we're building we can just select the proper file at container build time rather than manually changing it.

akien-mga:
I'll merge as is for now to move on for 3.2 beta 3, but we can improve further.
As the 3.1 builds are done from the `3.1` branch, and we likely won't want to go back to Mono 5 for 3.2+, it's probably fine that the OSX Dockerfile is tailor-made for 6.0+.

PR #26, opened 2019-11-26
Fixed Dockerfile::CMD syntax for Docker compatibility

dacrystal:
Fixed issue for docker to parse CMD properly.
`/bin/sh: [/bin/bash]: No such file or directory`

From [Dockerfile  Reference ](https://docs.docker.com/engine/reference/builder/) under **CMD** section: 
> Note: The exec form is parsed as a JSON array, which means that you must use double-quotes (“) around words not single-quotes (‘).

Tested  on :
```
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.1
BuildVersion:	19B88

$ docker --version
Docker version 19.03.4, build 9013bf5
```

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -10,4 +10,4 @@ RUN dnf -y install scons java-1.8.0-openjdk-devel ncurses-compat-libs unzip whic
 ENV ANDROID_HOME=/root/
 ENV ANDROID_NDK_ROOT=/root/ndk-bundle/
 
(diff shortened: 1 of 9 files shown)


akien-mga:
That's already fixed in the `master` branch with a5ceea36d, I prefer to keep both branches in sync as much as possible and thus not to merge custom changes for the `3.1` branch.

PR #29, opened 2019-12-04
Mono: Allow specifying fixed git hash and pretty image version

akien-mga:
Specifying only a branch like '2019-08' is problematic if upstream commits
something between the builds of two containers.

We could make a tarball in the first container build and reuse that, but we
need the flexibility of Git for some patching in the Android and JavaScript
builds, so instead we allow specifying a git tree-ish as the third argument.

We can therefore also give a pretty version string to use for the image name
as second argument, e.g. 6.6.0.160 for bef1e6335812d32f8eab648c0228fc624b9f8357.

As we can't `git clone` a commit hash directly, we still need to provide both
branch name and then commit hash for this use case.

Follow up to #23.

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -2,6 +2,7 @@ ARG img_version
 FROM godot-mono:${img_version}
 
 ARG mono_version
+ARG mono_commit
 
 RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
     dnf -y install --setopt=install_weak_deps=False \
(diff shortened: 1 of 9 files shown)


akien-mga:
With the above changes I get this nice set of images:
```
REPOSITORY                                            TAG                     IMAGE ID       CREATED             SIZE
localhost/godot-osx                                   master-mono-6.6.0.160   2588a2104f87   About an hour ago   2.65 GB
localhost/godot-ios                                   master-mono-6.6.0.160   6431c17eba79   2 hours ago         1.85 GB
localhost/godot-javascript                            master-mono-6.6.0.160   d03c10ea28e4   2 hours ago         2.74 GB
localhost/godot-android                               master-mono-6.6.0.160   ba83b69aaf2d   2 hours ago         13.3 GB
localhost/godot-ubuntu-32                             master-mono-6.6.0.160   54812d0f6159   3 hours ago         995 MB
localhost/godot-ubuntu-64                             master-mono-6.6.0.160   d15e671bc1f7   3 hours ago         1.07 GB
localhost/godot-windows                               master-mono-6.6.0.160   6b4c7b1d67b1   4 hours ago         2.12 GB
localhost/godot-mono-glue                             master-mono-6.6.0.160   30de478b7116   4 hours ago         1.51 GB
localhost/godot-mono                                  master-mono-6.6.0.160   21371aac3535   4 hours ago         1.26 GB
localhost/godot-export                                master-mono-6.6.0.160   951a62f0414c   4 hours ago         863 MB
localhost/godot-xcode-packer                          master-mono-6.6.0.160   ab1cb0e3c8f8   25 hours ago        982 MB
localhost/godot-fedora                                master-mono-6.6.0.160   067fec69f143   25 hours ago        438 MB
```

When running:
```
./build.sh master mono-6.6.0.160 2019-08 bef1e6335812d32f8eab648c0228fc624b9f8357
```

PR #35, opened 2020-01-07
Windows: Set WINEPATH for 32-bit wine to fix Mono build

### LINKED ISSUE #34: Windows 32-bit Mono build error with the Wine hack
While the 64-bit Windows Mono build works fine with `wine64`, we get an issue in the 32-bit Windows Mono build (using `wine32`).

The build itself runs fine, and the resulting container actually works, so I'm not sure exactly what step is raising the error, but it prompted the need for #33 to solve a subsequent issue with generating the symlinks.

The actual error happens during `make install`, where apparently some step involves running `mono-sgen.exe` again from the mini distrib, but it fails with:
```
Creating ../../build/deps/win32_net_4_x_corlib_test.dll.makefrag ...
Creating ../../build/deps/net_4_x_corlib_xunit-test.dll.makefrag ...
000b:fixme:winediag:start_process Wine Staging 4.21 is a testing version containing experimental patches.
000b:fixme:winediag:start_process Please mention your exact version when filing bug reports on winehq.org.
002a:err:module:import_dll Library libgcc_s_sjlj-1.dll (which is needed by L"Z:\\root\\mono\\mono\\mini\\.libs\\libmonosgen-2.0.dll") not found
002a:err:module:import_dll Library libmonosgen-2.0.dll (which is needed by L"Z:\\root\\mono\\mono\\mini\\.libs\\mono-sgen.exe") not found
002a:err:module:LdrInitializeThunk Importing dlls for L"Z:\\root\\mono\\mono\\mini\\.libs\\mono-sgen.exe" failed, status c0000135
make[8]: *** [il/il.make:4: ../../class/lib/net_4_x-win32/corlib.unsafe.dll.tmp] Error 53
make[8]: *** Waiting for unfinished jobs....
```

Before tweaking the Wine build script to allow using `wine32` instead of `wine` (64), I used to have a similar issue further up during `make` which would prevent the build altogether. So I guess for some reason it's trying again to run with `wine64`, or it's actually trying to run a 64-bit binary with `wine32` maybe?

CC @hpvb

akien-mga:
Symlinking the BCL from `/usr/lib/mono` no longer seems needed
for either architecture, everything is built properly.

Fixes #34.

----

This seemed to work in-container when doing a test. I'm now making a full rebuild of the container and will do a test Mono win64 and win32 build off the `master` branch to confirm that everything works as intended.

### DIFF
--- a/files/mono-build-win32.sh
+++ b/files/mono-build-win32.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 
-echo -e '#!/bin/bash\nwine${WINE_BITS} $(dirname $0)/mono-sgen.exe "$@"' > mono/mini/mono
+if [ "${WINE_BITS}" == "64" ]; then
+  export WINEPATH="/usr/x86_64-w64-mingw32/sys-root/mingw/bin/"
+else
+  export WINEPATH="/usr/i686-w64-mingw32/sys-root/mingw/bin/"
+fi
+
+echo -e '#!/bin/bash\n'"wine${WINE_BITS}"' $(dirname $0)/mono-sgen.exe "$@"' > mono/mini/mono
 chmod +x mono/mini/mono
 
 mkdir -p .bin
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -20,16 +20,14 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mand
     rm -f /root/dependencies/mono-64/bin/mono /root/dependencies/mono-64/bin/mono-sgen && \
     ln -s /usr/bin/mono /root/dependencies/mono-64/bin/mono && \
     ln -s /usr/bin/mono-sgen /root/dependencies/mono-64/bin/mono-sgen && \
-    (ln -sf /usr/lib/mono/* /root/dependencies/mono-64/lib/mono/ || /bin/true) && \
     cp -rvp /etc/mono /root/dependencies/mono-64/etc && \
     export WINE_BITS=32 && \
-    (bash /root/files/mono-build-win32.sh --prefix=/root/dependencies/mono-32 --host=i686-w64-mingw32 || /bin/true) && \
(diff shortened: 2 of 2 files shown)


files/mono-build-win32.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 
-echo -e '#!/bin/bash\nwine${WINE_BITS} $(dirname $0)/mono-sgen.exe "$@"' > mono/mini/mono
+if [ "${WINE_BITS}" == "64" ]; then
+  export WINEPATH="/usr/x86_64-w64-mingw32/sys-root/mingw/bin/"
+else
+  export WINEPATH="/usr/i686-w64-mingw32/sys-root/mingw/bin/"
+fi
+
+echo -e '#!/bin/bash\n'"wine${WINE_BITS}"' $(dirname $0)/mono-sgen.exe "$@"' > mono/mini/mono
```
akien-mga: `${WINE_BITS}` was not expanded due to the single quotes.

files/mono-build-win32.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 
-echo -e '#!/bin/bash\nwine${WINE_BITS} $(dirname $0)/mono-sgen.exe "$@"' > mono/mini/mono
+if [ "${WINE_BITS}" == "64" ]; then
+  export WINEPATH="/usr/x86_64-w64-mingw32/sys-root/mingw/bin/"
+else
+  export WINEPATH="/usr/i686-w64-mingw32/sys-root/mingw/bin/"
```
akien-mga: This was sufficient to fix the 32-bit build. I also added the 64-bit variant for consistency, just in case.

hpvb:
It may be better to set a WINEARCH and a WINEPREFIX for each build instead.

akien-mga:
> It may be better to set a WINEARCH and a WINEPREFIX for each build instead.

How would that look like? Would it be sufficient for wine to then properly pick up its base DLLs to load `mono-sgen.exe`?

akien-mga:
Merged by mistake when pushing another change to the `master` branch, oh well... it does work at least :)

PR #38, opened 2020-03-10
Fix windows docker, reduce redundancy, hadolint recommendations

### LINKED ISSUE #37: Dockerfile.Windows doesn't build
I started a new build of containers with `./build.sh 3.2 mono-6.6.0.166`. What version of mono is currently being used for the official 3.2 binaries?

It fails when building windows. Initially, it hangs for a very long time after `Cloning into '/root/mono/external/corefx'...` or on `corert`. I've run it several times over multiple days and it hangs in one of those two spots. I left it overnight and it finally moved past it.

```
Cloning into '/root/mono/external/roslyn-binaries'...
Cloning into '/root/mono/external/rx'...
Cloning into '/root/mono/external/xunit-binaries'...
Cloning into '/root/mono/external/corefx'...
Submodule path 'external/Newtonsoft.Json': checked out '471c3e0803a9f40a0acc8aeceb31de6ff93a52c4'
Submodule path 'external/api-doc-tools': checked out '5da8127af9e68c9d58a90aa9de21f57491d81261'
Submodule path 'external/api-snapshot': checked out '7753b257899d0d0e51f02b60847a5b632f11bbc3'
Submodule path 'external/aspnetwebstack': checked out 'e77b12e6cc5ed260a98447f609e887337e44e299'
```

Then it ultimately fails with:
```
/usr/lib/gcc/i686-w64-mingw32/9.2.1/../../../../i686-w64-mingw32/bin/as: error while loading shared libraries: libc.so.6: cannot stat shared
 object: Error 103
/usr/libexec/gcc/i686-w64-mingw32/9.2.1/cc1: error while loading shared libraries: libc.so.6: cannot stat shared object: Error 103
i686-w64-mingw32-gcc: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: Error 107
mv: error while loading shared libraries: libc.so.6: cannot stat shared object: Error 103
i686-w64-mingw32-gcc: error while loading shared libraries: libc.so.6: cannot stat shared object: Error 103
i686-w64-mingw32-gcc: error while loading shared libraries: libc.so.6: cannot stat shared object: Error 107
cat: error while loading shared libraries: libc.so.6: cannot stat shared object: Error 107
i686-w64-mingw32-gcc: error while loading shared libraries: libc.so.6: cannot stat shared object: Error 107
mv: error while loading shared libraries: libc.so.6: cannot stat shared object: Error 107
container exited on bus error
Error: error building at STEP "RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi &&     dnf -y install --setopt=install_weak_deps=False       mingw32-gcc mingw32-gcc-c++ mingw32-winpthreads-static mingw64-gcc mingw64-gcc-c++ mingw64-winpthreads-static yasm wine &&     dnf clean all &&     git clone https://github.com/mono/mono --branch ${mono_version} --single-branch &&     cd /root/mono &&     if [ ! -z "${mono_commit}" ]; then git checkout ${mono_commit}; fi &&     git submodule update --init &&     git apply -3 /root/files/patches/mono-unity-Clear-TLS-instead-of-aborting.patch &&     git apply -3 /root/files/patches/wine-mono.patch &&     export WINE_BITS=64 &&     bash /root/files/mono-build-win32.sh --prefix=/root/dependencies/mono-64 --host=x86_64-w64-mingw32 &&     git clean -fdx &&     cp /root/dependencies/mono-64/bin/libMonoPosixHelper.dll /root/dependencies/mono-64/bin/MonoPosixHelper.dll &&     rm -f /root/dependencies/mono-64/bin/mono /root/dependencies/mono-64/bin/mono-sgen &&     ln -s /usr/bin/mono /root/dependencies/mono-64/bin/mono &&     ln -s /usr/bin/mono-sgen /root/dependencies/mono-64/bin/mono-sgen &&     cp -rvp /etc/mono /root/dependencies/mono-64/etc &&     export WINE_BITS=32 &&     bash /root/files/mono-build-win32.sh --prefix=/root/dependencies/mono-32 --host=i686-w64-mingw32 &&     cd /root &&     cp /root/dependencies/mono-32/bin/libMonoPosixHelper.dll /root/dependencies/mono-32/bin/MonoPosixHelper.dll &&     rm -f /root/dependencies/mono-32/bin/mono /root/dependencies/mono-32/bin/mono-sgen &&     ln -s /usr/bin/mono /root/dependencies/mono-32/bin/mono &&     ln -s /usr/bin/mono-sgen /root/dependencies/mono-32/bin/mono-sgen &&     cp -rvp /etc/mono /root/dependencies/mono-32/etc &&     rm -rf /root/mono &&     dnf -y remove wine": error while running runtime: exit status 1

```

[windows.log](https://github.com/godotengine/build-containers/files/4302966/windows.log)

tinmanjuggernaut:
This PR fixes problems in Dockerfile.windows, and removes a lot of redundant downloading and processing.

Fixes #37 

* Reduces downloading mono and ~25 submodules from 7x to once (and saves on re-runs)
* Reduces dnf cache downloads from 11x to once (dnf cleans packages only, retaining cache)
* Reduces git clone/submodule downloads by downloading shallow repos where possible (`--depth 1`)
* Adds progress indicator to all git clones
* Fixes patching. `git apply` seems less reliable and will reject everything if anything is the tiniest bit off. Patch works in chunks and will patch what it can. It has more tolerance for changes, e.g. line number changes. 2 of 3 `git apply` commands failed. `patch -p1 < file` worked. Replaced all three patch commands.
* Moves building xcode-packer into `if ! -e macos sdk`. Build repo only if needed.
* Dockerfile.windows:
    * Fixes can't find libstdc++.so.6 and libc.so.6 (not sure how, but giving it a clean repo between bit versions helped a lot of failures)
   * Bash scripts and Make now print out every command so errors are easier to track
   * Fixes `dnf remove wine` hanging on avahi scriptlet
   * Fixes building Jay

I just built all new images. But why is android so big?! Now it's 5gb bigger than before!

```
localhost/godot-android        3.2-mono-6.6.0.166   fa87ed0c9068   3 minutes ago   17 GB
localhost/godot-xcode-packer   3.2-mono-6.6.0.166   a63d2620acdf   3 hours ago     1.03 GB
localhost/godot-ios            3.2-mono-6.6.0.166   0758942928de   17 hours ago    1.88 GB
localhost/godot-javascript     3.2-mono-6.6.0.166   fb34f8055402   19 hours ago    2.83 GB
localhost/godot-ubuntu-32      3.2-mono-6.6.0.166   f657b619d48d   20 hours ago    994 MB
localhost/godot-ubuntu-64      3.2-mono-6.6.0.166   3b16c95d9cee   20 hours ago    1.07 GB
localhost/godot-windows        3.2-mono-6.6.0.166   afc2f33e1efb   21 hours ago    3.45 GB
localhost/godot-mono-glue      3.2-mono-6.6.0.166   d3f36cbf40d2   23 hours ago    1.57 GB
localhost/godot-mono           3.2-mono-6.6.0.166   858d6311024b   23 hours ago    1.36 GB
localhost/godot-export         3.2-mono-6.6.0.166   174885334dde   24 hours ago    949 MB
localhost/godot-fedora         3.2-mono-6.6.0.166   1810aa32f22b   24 hours ago    564 MB
docker.io/library/fedora       31                   536f3995adeb   2 weeks ago     200 MB

---
Large android folders in /root:
8.9G    android-toolchains
1.5G    mono-configs
98M     mono-installs
4.1G    sdk

1.3G    /usr
```

--- 
Second commit, I ran http://github.com/hadolint/hadolint per calinou. This resulted in 45 best practices warnings. I fixed all except for using WORKDIR instead of cd, and pinning apt-get installs (and dnf) to specific versions).
```
DL3003 Use WORKDIR to switch to a directory
DL3008 Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
```

Key takeaways:
* Any docker that uses `|` should have `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` before RUN so it will fail if either side of the pipe fails.
* Globs `*.deb` should begin with `./*.deb` in case a file begins with a hyphen so it's not interpreted as an option.
* `-e` on `echo` is not a POSIX standard and printf is recommended.
* `CMD` is a json list, so should be formatted as `CMD ["/bin/bash"]`
* Because of the above, I combined xcode RUN and CMD into one command and removed the redundant line in build.sh

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -1,9 +1,9 @@
 ARG img_version
 FROM godot-mono:${img_version}
-
 ARG mono_version
 
-RUN if [ -z "${mono_version}" ]; then echo -e "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+RUN if [ -z "${mono_version}" ]; then printf "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
     dnf -y install --setopt=install_weak_deps=False \
       gcc gcc-c++ java-1.8.0-openjdk-devel ncurses-compat-libs && \
     dnf clean packages && \
@@ -32,4 +32,4 @@ RUN cp -a /root/files/${mono_version} /root && \
     cd /root && \
     rm -rf /root/${mono_version} /root/godot-mono-builds
 
-CMD /bin/bash
+CMD ["/bin/bash"]
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -2,12 +2,11 @@ ARG img_version
 FROM godot-mono:${img_version}
 
 ARG mono_version
-ARG mono_commit
 
-RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
+RUN if [ -z "${mono_version}" ]; then echo -e "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
     dnf -y install --setopt=install_weak_deps=False \
       gcc gcc-c++ java-1.8.0-openjdk-devel ncurses-compat-libs && \
-    dnf clean all && \
+    dnf clean packages && \
     mkdir sdk && cd sdk && \
     curl -LO https://dl.google.com/android/repository/sdk-tools-linux-4333796.zip && \
     unzip sdk-tools-linux-4333796.zip && \
@@ -18,12 +17,8 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mand
 ENV ANDROID_HOME=/root/sdk/
 ENV ANDROID_NDK_ROOT=/root/sdk/ndk-bundle/
 
-RUN git clone https://github.com/mono/mono --branch ${mono_version} --single-branch && \
-    cd /root/mono && \
-    if [ ! -z "${mono_commit}" ]; then git checkout ${mono_commit}; fi && \
-    git submodule update --init && \
-    git apply -3 /root/files/patches/mono-unity-Clear-TLS-instead-of-aborting.patch && \
-    export MONO_SOURCE_ROOT=/root/mono && \
+RUN cp -a /root/files/${mono_version} /root && \
+    export MONO_SOURCE_ROOT=/root/${mono_version} && \
     export make="make -j" && \
     git clone https://github.com/godotengine/godot-mono-builds /root/godot-mono-builds && \
     cd /root/godot-mono-builds && \
@@ -31,10 +26,10 @@ RUN git clone https://github.com/mono/mono --branch ${mono_version} --single-bra
     python3 patch_mono.py && \
     python3 android.py configure --target=all-runtime && \
     python3 android.py make --target=all-runtime && \
-    cd /root/mono && git clean -fdx && NOCONFIGURE=1 ./autogen.sh && \
+    cd /root/${mono_version} && git clean -fdx && NOCONFIGURE=1 ./autogen.sh && \
     cd /root/godot-mono-builds && \
     python3 bcl.py make --product=android && \
     cd /root && \
-    rm -rf /root/mono /root/godot-mono-builds
+    rm -rf /root/${mono_version} /root/godot-mono-builds
 
 CMD /bin/bash
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -7,4 +7,4 @@ RUN dnf -y upgrade --setopt=install_weak_deps=False && \
       bash bzip2 curl git make patch pkgconfig python3 scons unzip which xz && \
     dnf clean packages
 
-CMD /bin/true
+CMD ["/bin/bash"]
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -5,6 +5,6 @@ WORKDIR /root
 RUN dnf -y upgrade --setopt=install_weak_deps=False && \
     dnf -y install --setopt=install_weak_deps=False \
       bash bzip2 curl git make patch pkgconfig python3 scons unzip which xz && \
-    dnf clean all
+    dnf clean packages
 
 CMD /bin/true
--- a/Dockerfile.javascript
+++ b/Dockerfile.javascript
@@ -2,36 +2,32 @@ ARG img_version
 FROM godot-mono:${img_version}
 
 ARG mono_version
-ARG mono_commit
 
 RUN dnf -y install --setopt=install_weak_deps=False \
       java-openjdk yasm && \
-    dnf clean all && \
-    git clone https://github.com/emscripten-core/emsdk && \
+    dnf clean packages && \
+    git clone https://github.com/emscripten-core/emsdk --progress && \
     cd emsdk && \
     git checkout a5082b232617c762cb65832429f896c838df2483 && \
     ./emsdk install 1.38.47-upstream && \
     ./emsdk activate 1.38.47-upstream && \
     echo "source /root/emsdk/emsdk_env.sh" >> /root/.bashrc
 
-RUN git clone https://github.com/mono/mono --branch ${mono_version} --single-branch && \
-    cd /root/mono && \
-    if [ ! -z "${mono_commit}" ]; then git checkout ${mono_commit}; fi && \
-    git submodule update --init && \
-    git apply -3 /root/files/patches/mono-unity-Clear-TLS-instead-of-aborting.patch && \
-    git apply -3 /root/files/patches/mono-pr16636-wasm-bugfix-and-update.diff && \
-    export MONO_SOURCE_ROOT=/root/mono && \
+RUN cp -a /root/files/${mono_version} /root && \
+    cd /root/${mono_version} && \
+    patch -p1 < /root/files/patches/mono-pr16636-wasm-bugfix-and-update.diff && \
+    export MONO_SOURCE_ROOT=/root/${mono_version} && \
     export make="make -j" && \
-    git clone https://github.com/godotengine/godot-mono-builds /root/godot-mono-builds && \
+    git clone https://github.com/godotengine/godot-mono-builds /root/godot-mono-builds --progress && \
     cd /root/godot-mono-builds && \
     git checkout 710b275fbf29ddb03fd5285c51782951a110cdab && \
     python3 patch_emscripten.py && \
     python3 wasm.py configure --target=runtime && \
     python3 wasm.py make --target=runtime && \
-    cd /root/mono && git clean -fdx && NOCONFIGURE=1 ./autogen.sh && \
+    cd /root/${mono_version} && git clean -fdx && NOCONFIGURE=1 ./autogen.sh && \
     cd /root/godot-mono-builds && \
     python3 bcl.py make --product wasm && \
     cd /root && \
-    rm -rf /root/mono /root/godot-mono-builds
+    rm -rf /root/${mono_version} /root/godot-mono-builds
 
 CMD /bin/bash
--- a/Dockerfile.javascript
+++ b/Dockerfile.javascript
@@ -3,7 +3,8 @@ FROM godot-mono:${img_version}
 
 ARG mono_version
 
-RUN dnf -y install --setopt=install_weak_deps=False \
+RUN if [ -z "${mono_version}" ]; then printf "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
+    dnf -y install --setopt=install_weak_deps=False \
       java-openjdk yasm && \
     dnf clean packages && \
     git clone https://github.com/emscripten-core/emsdk --progress && \
@@ -30,4 +31,4 @@ RUN cp -a /root/files/${mono_version} /root && \
     cd /root && \
     rm -rf /root/${mono_version} /root/godot-mono-builds
 
-CMD /bin/bash
+CMD ["/bin/bash"]
--- a/Dockerfile.ubuntu-32
+++ b/Dockerfile.ubuntu-32
@@ -2,9 +2,11 @@ FROM i386/ubuntu:trusty
 
 ARG mono_version
 
-RUN if [ -z "${mono_version}" ]; then echo -e "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+RUN if [ -z "${mono_version}" ]; then printf "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
     apt-get update && \
-    apt-get -y install wget && \
+    apt-get -y install wget --no-install-recommends && \
     cd /root && \
     wget -O- 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x1E9377A2BA9EF27F' | apt-key add - && \
     wget -O- 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8E51A6D660CD88D67D65221D90BD7EACED8E640A' | apt-key add - && \
@@ -18,7 +20,9 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nArgument mono_version is mand
     ln -sf /usr/bin/gcc-ranlib-8 /usr/bin/gcc-ranlib && \
     ln -sf /usr/bin/gcc-ar-8 /usr/bin/gcc-ar && \
     ln -sf /usr/bin/gcc-8 /usr/bin/gcc && \
-    ln -sf /usr/bin/g++-8 /usr/bin/g++
+    ln -sf /usr/bin/g++-8 /usr/bin/g++ && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 
 RUN cp -a /root/files/${mono_version} /root && \
     cd /root/${mono_version} && \
@@ -33,11 +37,9 @@ RUN cp -a /root/files/${mono_version} /root && \
     wget https://download.mono-project.com/repo/ubuntu/pool/main/c/core-setup/msbuild-libhostfxr_3.0.0.2019.04.16.02.13-0xamarin3+ubuntu1604b1_i386.deb && \
     wget https://download.mono-project.com/repo/ubuntu/pool/main/m/msbuild/msbuild-sdkresolver_16.3+xamarinxplat.2019.08.08.00.55-0xamarin2+ubuntu1604b1_all.deb && \
     wget https://download.mono-project.com/repo/ubuntu/pool/main/n/nuget/nuget_5.2.0.6090.bin-0xamarin1+ubuntu1604b1_all.deb && \
-    dpkg -i --force-all *.deb && \
+    dpkg -i --force-all ./*.deb && \
     sed -i '/Depends.*mono/d' /var/lib/dpkg/status && \
     ln -s /usr/bin/mono /usr/bin/cli && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/ && \
-    rm *.deb
+    rm ./*.deb
 
-CMD /bin/bash
+CMD ["/bin/bash"]
--- a/Dockerfile.ubuntu-32
+++ b/Dockerfile.ubuntu-32
@@ -1,9 +1,8 @@
 FROM i386/ubuntu:trusty
 
 ARG mono_version
-ARG mono_commit
 
-RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
+RUN if [ -z "${mono_version}" ]; then echo -e "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
     apt-get update && \
     apt-get -y install wget && \
     cd /root && \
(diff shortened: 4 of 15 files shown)


tinmanjuggernaut:
Ok, this is ready for review. It built all dockers, except I haven't setup UWP yet. I was able to build Godot with all of them including android/ios. 

However:

* The Windows/Mono binary won't run because [mono_configure.py](https://github.com/godotengine/godot/blob/master/modules/mono/build_scripts/mono_configure.py) doesn't copy the needed mono files (e.g. mscorlib.dll) into the GodotSharp directory. It copies them fine for the linux install (same Godot source tree for both builds). Already documented here: https://github.com/godotengine/godot/issues/34825

* I had to add an extra include flag into one of the android/ios builds to get it to compile. I'm in windows- I'll update this line once I reboot and can get that info....

Dockerfile.base
```diff
@@ -5,6 +5,6 @@ WORKDIR /root
 RUN dnf -y upgrade --setopt=install_weak_deps=False && \
     dnf -y install --setopt=install_weak_deps=False \
       bash bzip2 curl git make patch pkgconfig python3 scons unzip which xz && \
-    dnf clean all
+    dnf clean packages
```
tinmanjuggernaut: Don't delete the lists that just took 10min to download. Save them for all the other dockers rather than wasting that time over and over.

Dockerfile.javascript
```diff
@@ -2,36 +2,32 @@ ARG img_version
...
-RUN git clone https://github.com/mono/mono --branch ${mono_version} --single-branch && \
-    cd /root/mono && \
-    if [ ! -z "${mono_commit}" ]; then git checkout ${mono_commit}; fi && \
-    git submodule update --init && \
-    git apply -3 /root/files/patches/mono-unity-Clear-TLS-instead-of-aborting.patch && \
-    git apply -3 /root/files/patches/mono-pr16636-wasm-bugfix-and-update.diff && \
-    export MONO_SOURCE_ROOT=/root/mono && \
+RUN cp -a /root/files/${mono_version} /root && \
+    cd /root/${mono_version} && \
+    patch -p1 < /root/files/patches/mono-pr16636-wasm-bugfix-and-update.diff && \
```
tinmanjuggernaut: patch seems more reliable than git apply.

Dockerfile.windows
```diff
@@ -2,35 +2,37 @@ ARG img_version
...
+    bash -x /root/files/mono-build-win32.sh --prefix=/root/dependencies/mono-32 --host=i686-w64-mingw32 && \
     cd /root && \
+    rm -rf /root/${mono_version} && \
     cp /root/dependencies/mono-32/bin/libMonoPosixHelper.dll /root/dependencies/mono-32/bin/MonoPosixHelper.dll && \
     rm -f /root/dependencies/mono-32/bin/mono /root/dependencies/mono-32/bin/mono-sgen && \
     ln -s /usr/bin/mono /root/dependencies/mono-32/bin/mono && \
     ln -s /usr/bin/mono-sgen /root/dependencies/mono-32/bin/mono-sgen && \
     cp -rvp /etc/mono /root/dependencies/mono-32/etc && \
-    rm -rf /root/mono && \
+    rpm -ev --nodeps --noscripts --notriggers avahi && \
```
tinmanjuggernaut: Current version of avahi has a broken uninstall script that just calls systemd, which of course isn't running. The whole build will hang trying to uninstall this automatically. Going through the docker manually, dnf will hang. Only way is to manually remove it without running the scripts. This line can go away if they every fix it.

Dockerfile.windows
```diff
@@ -2,35 +2,37 @@ ARG img_version
...
+    dnf clean packages && \
+    cp -a /root/files/${mono_version} /root && \
+    cd /root/${mono_version} && \
+    patch -p1 < /root/files/patches/wine-mono.patch && \
     export WINE_BITS=64 && \
-    bash /root/files/mono-build-win32.sh --prefix=/root/dependencies/mono-64 --host=x86_64-w64-mingw32 && \
-    git clean -fdx && \
+    bash -x /root/files/mono-build-win32.sh --prefix=/root/dependencies/mono-64 --host=x86_64-w64-mingw32 && \
+    cd /root && \
+    rm -rf /root/${mono_version} && \
```
tinmanjuggernaut: Removing this and having a completely clean mono directory between bit versions seemed to work better than just make clean. At least on 6.6.0.166. I had difficulties until I did this.

build.sh
```diff
@@ -1,38 +1,63 @@
...
-echo "Mono version used: ${mono_version} ${mono_commit}"
+
+# If mono branch does not start with mono-, prepend it to the folder name.
+if [ ${mono_version:0:5} != "mono-" ]; then
+
+	mono_root=${files_root}/"mono-${mono_version}"
+else
+	mono_root=${files_root}/${mono_version}
+fi
+
```
tinmanjuggernaut: So if you request the mono-6.6.0.166 branch, that's what the folder is called.
If you request the 2019-08 branch, the folder is mono-2019-08.

build.sh
```diff
@@ -45,21 +70,32 @@ done
...
+    git clone https://github.com/mono/mono ${mono_root} --branch ${mono_version} --single-branch --progress --depth 1 
+    pushd ${mono_root}
+    if [ ! -z "${mono_commit}" ]; then 
+	git checkout ${mono_commit} 
+    fi
+    git submodule update --init --recursive --recommend-shallow --progress --depth 1 
+    patch -p1 < ${files_root}/patches/mono-unity-Clear-TLS-instead-of-aborting.patch 
+    popd
+fi
+
```
tinmanjuggernaut: Download mono plus 25+ submodules once instead of 7x! Also only need shallow depth, but I don't think it saved much.

files/mono-build-win32.sh
```diff
@@ -17,13 +17,13 @@ export PATH="$(pwd)/.bin/:$PATH"
 ./autogen.sh $@ --disable-boehm --with-mcs-docs=no HOST_PROFILE=win32
 echo '#define HAVE_STRUCT_SOCKADDR_IN6 1' >> config.h
 pushd mcs/jay
-make CC=gcc
+make CC=gcc BUILD_PLATFORM=win32
```
tinmanjuggernaut: Fix jay compilation error noted in #37

Dockerfile.android
```diff
@@ -1,9 +1,9 @@
 ARG img_version
 FROM godot-mono:${img_version}
-
 ARG mono_version
 
-RUN if [ -z "${mono_version}" ]; then echo -e "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
tinmanjuggernaut: hadolint says use this line for any run statement that has a pipe in it so the docker will fail if either side of the pipe fails.

Dockerfile.android
```diff
@@ -1,9 +1,9 @@
...
+RUN if [ -z "${mono_version}" ]; then printf "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
```
tinmanjuggernaut: -e on `echo -e` is not a posix standard. hadolint says use printf.

Dockerfile.android
```diff
@@ -32,4 +32,4 @@ RUN cp -a /root/files/${mono_version} /root && \
     cd /root && \
     rm -rf /root/${mono_version} /root/godot-mono-builds
 
-CMD /bin/bash
+CMD ["/bin/bash"]
```
tinmanjuggernaut: CMD is supposed to be json formatted

Dockerfile.ubuntu-32
```diff
@@ -18,7 +20,9 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nArgument mono_version is mand
     ln -sf /usr/bin/gcc-ranlib-8 /usr/bin/gcc-ranlib && \
     ln -sf /usr/bin/gcc-ar-8 /usr/bin/gcc-ar && \
     ln -sf /usr/bin/gcc-8 /usr/bin/gcc && \
-    ln -sf /usr/bin/g++-8 /usr/bin/g++
+    ln -sf /usr/bin/g++-8 /usr/bin/g++ && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
```
tinmanjuggernaut: clean in this layer, not the second layer when the changes are permanent.

Dockerfile.ubuntu-32
```diff
@@ -33,11 +37,9 @@ RUN cp -a /root/files/${mono_version} /root && \
...
     wget https://download.mono-project.com/repo/ubuntu/pool/main/m/msbuild/msbuild-sdkresolver_16.3+xamarinxplat.2019.08.08.00.55-0xamarin2+ubuntu1604b1_all.deb && \
     wget https://download.mono-project.com/repo/ubuntu/pool/main/n/nuget/nuget_5.2.0.6090.bin-0xamarin1+ubuntu1604b1_all.deb && \
-    dpkg -i --force-all *.deb && \
+    dpkg -i --force-all ./*.deb && \
     sed -i '/Depends.*mono/d' /var/lib/dpkg/status && \
     ln -s /usr/bin/mono /usr/bin/cli && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/ && \
-    rm *.deb
+    rm ./*.deb
```
tinmanjuggernaut: add ./ before globs in case a filename beginning with - is interpreted as an option

Dockerfile.xcode
```diff
@@ -12,9 +14,8 @@ RUN dnf -y install --setopt=install_weak_deps=False \
     git clone https://github.com/NiklasRosenstein/pbzx --progress && \
     cd pbzx && \
     git checkout 2a4d7c3300c826d918def713a24d25c237c8ed53 && \
-    clang -O3 -llzma -lxar -I /usr/local/include pbzx.c -o pbzx
-
-CMD mkdir -p /root/xcode && \
+    clang -O3 -llzma -lxar -I /usr/local/include pbzx.c -o pbzx && \
+    mkdir -p /root/xcode && \
```
tinmanjuggernaut: CMD is json formatted. Rather than continue this separate RUN/CMD and putting in all those commands into json format, I wrapped it all up into RUN. We don't need separate docker layers for the two steps do we?

build.sh
```diff
@@ -105,7 +105,6 @@ if [ ! -e files/MacOSX10.14.sdk.tar.xz ] || [ ! -e files/iPhoneOS12.4.sdk.tar.xz
 
   echo "Building OSX and iOS SDK packages. This will take a while"
   $podman_build -t godot-xcode-packer:${img_version} -f Dockerfile.xcode -v ${files_root}:/root/files . 2>&1 | tee logs/xcode.log
-  $podman run -it --rm -v ${files_root}:/root/files godot-xcode-packer:${img_version} 2>&1 | tee logs/xcode_packer.log
```
tinmanjuggernaut: Unnecessary due to Docker.xcode change

akien-mga:
That's a lot of unrelated changes which are pretty difficult to review. This repository is used in production for our official binaries, so I can't merge a mix of linter fixes and actual build changes that might impact the containers.

I would suggest making separate PRs for unrelated changes so that they can be reviewed and merged individually.

akien-mga:
> I just built all new images. But why is android so big?! Now it's 5gb bigger than before!

When was "before"? It's big because it includes all the toolchains for Mono, which are ~10.5 GB in your `du` output.

Dockerfile.android
akien-mga: That's what it used to be, but containers failed running unless you specified `bash` manually AFAIR.

akien-mga: https://github.com/godotengine/build-containers/commit/a5ceea36da322bb7a41e7e714a88c8b46eccb412

Dockerfile.base
akien-mga: Fair enough, but 10 min sounds really excessive to download DNF metadata. Do you have a particularly slow connection/mirror?

Dockerfile.javascript
akien-mga: Alright.

tinmanjuggernaut:
> Separate PRs

fair enough

> Before

November-December my android build was only 12GB.

Dockerfile.base
tinmanjuggernaut: I've been in 3 countries in asia in the last several months, and about 7 living spaces. It varies wildly, but I find that the fedora servers are very slow even if I have a good connection. Right now I have 10-20mbps, but dnf downloads at 20-80kbps.

akien-mga:
> November-December my android build was only 12GB.

Was it before I merged Mono support in #7? Otherwise I'm not sure what would explain the size difference.

tinmanjuggernaut:
At that time I was building with mono 5.18.1.3, on Nov 12th, so it would not have included that patch.

build.sh
akien-mga: Makes sense. @hpvb's idea was to use a build stage for Mono instead to reuse the same stage in all containers without making them too big (which would happen if we closed Mono in the base container and just kept it there for downstream builds). But we can still do that later on, this is a good stopgap solution.

akien-mga: Does `--depth 1` still allow us to checkout a commit which might be older than `HEAD` though?

tinmanjuggernaut: No. You're right, that will break it. I had to remove that from the other dockers that checkout. Will remove that.

akien-mga:
> At that time I was building with mono 5.18.1.3, on Nov 12th, so it would not have included that patch.

So that's why, as you can see the Mono toolchain is heavy:
```
8.9G    android-toolchains
1.5G    mono-configs
98M     mono-installs
```
It includes binaries for all 4 architectures.
There might be ways to slim it down if need be, but that's to discuss on https://github.com/godotengine/godot-mono-builds

> I would suggest making separate PRs for unrelated changes so that they can be reviewed and merged individually.

Having had a quick look, I'd suggest:

- Move the second cosmetic commit to a separate PR. It can stay as is in that other PR with all linter fixes.
- The first commit could be split into several commits:
  * Move mono clone outside containers
  * Add `--progress` flag and change dnf cleanup to `packages` (could also be in the Mono move commit if we don't want to be overly pedantic :))
  * Fixes needed for Windows build (in a separate PR, as they need more review than the rest, and the Mono move could be merged first)

tinmanjuggernaut:
Sounds like a good plan. Thanks for the review and laying it out.

Dockerfile.base
akien-mga: I've noticed that Internet connectivity from containers (at least ours) can be a bit iffy -- which also explains your stalls when cloning Mono, which I also experienced a few times. So that might be why.
Though it's true that even on bare metal, Fedora metadata updates is always an unnecessary slow step.

Dockerfile.windows
akien-mga: Yeah `make clean` on Mono is simply broken. `git clean -fxd` should have worked fine though, but using a clean copy also works fine for me.

Dockerfile.windows
```diff
@@ -2,35 +2,37 @@ ARG img_version
...
-    git apply -3 /root/files/patches/mono-unity-Clear-TLS-instead-of-aborting.patch && \
-    git apply -3 /root/files/patches/wine-mono.patch && \
+    dnf clean packages && \
+    cp -a /root/files/${mono_version} /root && \
+    cd /root/${mono_version} && \
+    patch -p1 < /root/files/patches/wine-mono.patch && \
     export WINE_BITS=64 && \
-    bash /root/files/mono-build-win32.sh --prefix=/root/dependencies/mono-64 --host=x86_64-w64-mingw32 && \
-    git clean -fdx && \
+    bash -x /root/files/mono-build-win32.sh --prefix=/root/dependencies/mono-64 --host=x86_64-w64-mingw32 && \
```
akien-mga: Is `bash -x` a recent addition? I don't see it in `man bash` for bash 4.4.23 that I have locally.

tinmanjuggernaut: Same thing as `sh -x`. GNU Bash 5.0.3 (ubuntu) has it in the manual.

Dockerfile.windows
akien-mga: I never experienced this when building official containers, that's weird.
There hasn't been any avahi update in Fedora 31 since my previous builds in January either: https://src.fedoraproject.org/rpms/avahi/commits/f31

Dockerfile.windows
akien-mga: Well yes since `sh` is a symlink to `bash` ;) I guess you meant `set -x`.
It does seem to work with bash 4.4.23 too, I guess it's just the man page which is outdated. In any case it should work on F31 without issue.

build.sh
akien-mga: The `[ ! -e ${mono_root} ]` will also prevent checking out a different commit in the same branch, since only the branch name is part of the `mono_root`.

build.sh
```diff
@@ -45,21 +70,32 @@ done
...
-export podman_build_mono="$podman_build --build-arg mono_version=${mono_version} --build-arg mono_commit=${mono_commit} -v $(pwd)/files:/root/files"
+export podman_build_mono="$podman_build --build-arg mono_version=${mono_version} -v ${files_root}:/root/files"
 
 $podman build -t godot-fedora:${img_version} -f Dockerfile.base . 2>&1 | tee logs/base.log
 $podman_build -t godot-export:${img_version} -f Dockerfile.export . 2>&1 | tee logs/export.log
 
 $podman_build_mono -t godot-mono:${img_version} -f Dockerfile.mono . 2>&1 | tee logs/mono.log
 $podman_build_mono -t godot-mono-glue:${img_version} -f Dockerfile.mono-glue . 2>&1 | tee logs/mono-glue.log
 $podman_build_mono -t godot-windows:${img_version} -f Dockerfile.windows --ulimit nofile=65536 . 2>&1 | tee logs/windows.log
+
```
akien-mga: That's a nitpick but there's no real reason for having an empty line here and not between other builds ;)

files/mono-build-win32.sh
```diff
@@ -17,13 +17,13 @@ export PATH="$(pwd)/.bin/:$PATH"
...
 
 for dir in external/roslyn-binaries/Microsoft.Net.Compilers/[0-9]*; do
   MONO_PATH="$(winepath -w $(pwd)/${dir});${MONO_PATH}"
 done
 export MONO_PATH
 
-make -j
-make install
+make -j SHELL='bash -x'
+make install SHELL='bash -x'
```
akien-mga: Isn't `make` verbose enough as is?

files/mono-build-win32.sh
akien-mga: I can't reproduce the issue :(

files/mono-build-win32.sh
tinmanjuggernaut: I'll look again, but when I was debugging this, it was just stopping with an error, but no command. It might be fine without.

PR #40, opened 2020-03-19
Optimize mono & dnf downloads

tinmanjuggernaut:
* Moves mono downloads from dockers to build.sh
* Retains dnf cache
* Adds progress to git downloads
* Replaces `git apply` with more reliable `patch`

Should address relevant notes in reference #38 

Note: `dnf clean all` became `dnf clean packages` to save dnf cache and only remove the packages. However the [default dnf configuration](https://dnf.readthedocs.io/en/latest/conf_ref.html) doesn't save the packages. keepcache=false is the default and must be set true in `/etc/dnf/dnf.con` in order to save packages. So `dnf clean` is removed entirely everywhere.

### DIFF
--- a/Dockerfile.osx
+++ b/Dockerfile.osx
@@ -2,13 +2,11 @@ ARG img_version
 FROM godot-mono:${img_version}
 
 ARG mono_version
-ARG mono_commit
 
 RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
     dnf -y install --setopt=install_weak_deps=False \
       automake autoconf bzip2-devel clang libicu-devel libtool libxml2-devel llvm-devel openssl-devel yasm && \
-    dnf clean all && \
-    git clone https://github.com/tpoechtrager/osxcross.git && \
+    git clone --progress https://github.com/tpoechtrager/osxcross.git && \
     cd /root/osxcross && \
     git checkout 542acc2ef6c21aeb3f109c03748b1015a71fed63 && \
     ln -s /root/files/MacOSX10.14.sdk.tar.xz /root/osxcross/tarballs && \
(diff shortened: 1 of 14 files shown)


Dockerfile.osx
```diff
@@ -46,8 +41,6 @@ RUN git clone https://github.com/mono/mono --branch ${mono_version} --single-bra
     mkdir -p /root/dependencies/mono/etc && \
     cp -rvp /etc/mono /root/dependencies/mono/etc/ && \
     cp /root/files/mono-config-macosx /root/dependencies/mono/etc/mono/config && \
-    rm -rf /root/mono
-
-ENV MONO64_PREFIX=/root/dependencies/mono
+    rm -rf /root/${mono_version}
```
tinmanjuggernaut: deprecated and ignored by scons

Dockerfile.windows
```diff
@@ -25,15 +20,12 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mand
...
     rm -f /root/dependencies/mono-32/bin/mono /root/dependencies/mono-32/bin/mono-sgen && \
     ln -s /usr/bin/mono /root/dependencies/mono-32/bin/mono && \
     ln -s /usr/bin/mono-sgen /root/dependencies/mono-32/bin/mono-sgen && \
     cp -rvp /etc/mono /root/dependencies/mono-32/etc && \
-    rm -rf /root/mono && \
     dnf -y remove wine
 
-ENV MONO32_PREFIX=/root/dependencies/mono-32
-ENV MONO64_PREFIX=/root/dependencies/mono-64
-
```
tinmanjuggernaut: MONO32/64_PREFIX deprecated and ignored by scons

build.sh
```diff
@@ -1,65 +1,110 @@
...
-        [Yy]* ) break;;
-        [Nn]* ) exit 1;;
-        * ) echo "Please answer yes or no.";;
-    esac
+  read -p "Is this correct? [y/n] " yn
+  case $yn in
+    [Yy]* ) break;;
+    [Nn]* ) exit 1;;
+    * ) echo "Please answer yes or no.";;
+  esac
```
tinmanjuggernaut: Now `./build.sh master mono-6.6.0.160 2019-08 bef1e6335812d32f8eab648c0228fc624b9f8357` will make (and check for) a folder called `files/mono-2019-08-bef1e6`. Also it will say (exists) if found, so you know it won't be downloaded again.

build.sh
```diff
@@ -1,65 +1,110 @@
...
+    # Otherwise, get a shallow repo
+    git clone -b ${mono_version} --single-branch --progress --depth 1 https://github.com/mono/mono ${mono_root}
+    pushd ${mono_root}
+  fi
+  # Download all submodules, up to 6 at a time
+  git submodule update --init --recursive --recommend-shallow -j 6 --progress
+  patch -p1 < ${files_root}/patches/mono-unity-Clear-TLS-instead-of-aborting.patch 
+  popd
+fi
+
```
tinmanjuggernaut: `-j 6` made a huge difference in total download time. Less timeout and waiting because it was downloading other modules while one was timing out.

akien-mga:
Thanks!

PR #41, opened 2020-03-19
[WIP] Hadolint recommendations

TokisanGames:
I ran http://github.com/hadolint/hadolint per calinou. This resulted in 45 best practices warnings. I fixed all except for using WORKDIR instead of cd, and pinning apt-get installs (and dnf) to specific versions).
```
DL3003 Use WORKDIR to switch to a directory
DL3008 Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
```

Key takeaways:
* Any docker that uses `|` should have `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` before RUN so it will fail if either side of the pipe fails.
* Globs `*.deb` should begin with `./*.deb` in case a file begins with a hyphen so it's not interpreted as an option.
* `-e` on `echo` is not a POSIX standard and printf is recommended.
* `CMD` is a json list, so should be formatted as `CMD ["/bin/bash"]`
* Because of the above, I combined xcode RUN and CMD into one command and removed the redundant line in build.sh

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -2,12 +2,10 @@ ARG img_version
 FROM godot-mono:${img_version}
 
 ARG mono_version
-ARG mono_commit
 
 RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
     dnf -y install --setopt=install_weak_deps=False \
       gcc gcc-c++ java-1.8.0-openjdk-devel ncurses-compat-libs && \
-    dnf clean all && \
     mkdir sdk && cd sdk && \
     curl -LO https://dl.google.com/android/repository/sdk-tools-linux-4333796.zip && \
     unzip sdk-tools-linux-4333796.zip && \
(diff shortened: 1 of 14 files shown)


Dockerfile.xcode
```diff
@@ -11,9 +13,8 @@ RUN dnf -y install --setopt=install_weak_deps=False \
     git clone --progress https://github.com/NiklasRosenstein/pbzx && \
     cd pbzx && \
     git checkout 2a4d7c3300c826d918def713a24d25c237c8ed53 && \
-    clang -O3 -llzma -lxar -I /usr/local/include pbzx.c -o pbzx
-
-CMD mkdir -p /root/xcode && \
+    clang -O3 -llzma -lxar -I /usr/local/include pbzx.c -o pbzx && \
+    mkdir -p /root/xcode && \
```
tinmanjuggernaut: Merge into run command. Separate CMD w/ json format

Dockerfile.ubuntu-32
```diff
@@ -18,7 +20,10 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mand
     ln -sf /usr/bin/gcc-ranlib-8 /usr/bin/gcc-ranlib && \
     ln -sf /usr/bin/gcc-ar-8 /usr/bin/gcc-ar && \
     ln -sf /usr/bin/gcc-8 /usr/bin/gcc && \
-    ln -sf /usr/bin/g++-8 /usr/bin/g++
+    ln -sf /usr/bin/g++-8 /usr/bin/g++ && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/ && \
+    rm ./*.deb
```
tinmanjuggernaut: clean up files on the layer that installs them

Dockerfile.ubuntu-32
```diff
@@ -33,11 +38,8 @@ RUN cp -a /root/files/${mono_version} /root && \
     wget https://download.mono-project.com/repo/ubuntu/pool/main/c/core-setup/msbuild-libhostfxr_3.0.0.2019.04.16.02.13-0xamarin3+ubuntu1604b1_i386.deb && \
     wget https://download.mono-project.com/repo/ubuntu/pool/main/m/msbuild/msbuild-sdkresolver_16.3+xamarinxplat.2019.08.08.00.55-0xamarin2+ubuntu1604b1_all.deb && \
     wget https://download.mono-project.com/repo/ubuntu/pool/main/n/nuget/nuget_5.2.0.6090.bin-0xamarin1+ubuntu1604b1_all.deb && \
-    dpkg -i --force-all *.deb && \
+    dpkg -i --force-all ./*.deb && \
```
tinmanjuggernaut: avoid files that begin with `-` from being interpreted as an option

Dockerfile.android
```diff
@@ -3,7 +3,9 @@ FROM godot-mono:${img_version}
 
 ARG mono_version
 
-RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
tinmanjuggernaut: script stops if either side of a pipe fail

tinmanjuggernaut:
In spite of what hadolint says, `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` is apparently not accepted by podman. Though it doesn't cause any errors building the docker, it does report this message:
```
time="2020-03-25T15:53:41+08:00" level=error msg="SHELL is not supported for OCI image format, [/bin/bash -o pipefail -c] will be ignored. Must use `docker` format"
```

Making this one WIP while I figure it out.

akien-mga:
This should be rebased to remove the commits from #40 (merged) and #42 (not merged yet but I will after branching the `3.2` buildsystem off).

PR #42, opened 2020-03-25
Move Ubuntu to gcc 9.2.1

tinmanjuggernaut:
This merges against master.

I just built godot with both 32 and 64 bit dockers after this. There are few changed dependencies upon installing gcc 9. glibc is still 2.19.

Should not conflict with #40, but it will with #41 .

### DIFF
--- a/Dockerfile.ubuntu-32
+++ b/Dockerfile.ubuntu-32
@@ -14,12 +14,12 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mand
     apt-get -y update && \
     apt-get -y install --no-install-recommends \
       autoconf automake bzip2 cmake curl gettext git libtool make perl scons xz-utils \
-      gcc-8 g++-8 libudev-dev libx11-dev libxcursor-dev libxrandr-dev libasound2-dev libpulse-dev \
+      gcc-9 g++-9 libudev-dev libx11-dev libxcursor-dev libxrandr-dev libasound2-dev libpulse-dev \
       libgl1-mesa-dev libglu1-mesa-dev libxi-dev libxinerama-dev yasm && \
-    ln -sf /usr/bin/gcc-ranlib-8 /usr/bin/gcc-ranlib && \
-    ln -sf /usr/bin/gcc-ar-8 /usr/bin/gcc-ar && \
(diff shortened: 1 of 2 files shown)


akien-mga:
Thanks!

akien-mga:
Note: I branched off the `master` branch from before this merge to `3.2`, as the GCC version should stay pinned there for 3.2.x builds.

akien-mga:
Actually I changed my mind and reverted this PR for now (well, force pushed `master` without this commit), as it was a hassle to have to handle syncing `master` and `3.2` once they have diverged with this commit, since I have a lot of work to do for 3.2 which is also needed for master (iOS C# support, Android and WebAssembly AOT compilation support for C#).

I'll re-apply this patch once I'm done with the 3.2+master common part, and once I start working on the master branch for 4.0 (we're still a long way away from the first alpha build).

tinmanjuggernaut:
This was needed to build GD 4 previously, but now other changed requirements like scons 3 has rendered this docker unable to build GD 4 at all. I'm building 4 locally. We need to get off of trusty and move at least to 16.x, but that also bumps up our libc6 quite a bit.

akien-mga:
Well we can still use Ubuntu 14.04, having Python 3 support in SCons is just a matter of changing the bogus shebang to point to `/usr/bin/env python3` instead of `/usr/bin/env python`.

Then SCons requires Python 3.5 and Ubuntu 14.04 provides 3.4 by default, but it shouldn't be too hard to install 3.5 in parallel (I think I've seen instructions to do that).

akien-mga:
This should give a functional scons in Ubuntu 14.04:
```
apt-get update
apt-get remove scons
apt-get install python3.5
curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
python3.5 get-pip.py
pip install scons

# Then:
python3.5 /usr/local/bin/scons
# Or:
sed -i /usr/local/bin/scons -e 's,/usr/bin/env python$,/usr/bin/env python3.5,'
scons
```

Installing `pip` via `get-pip.py` (since there's no `python3.5-pip`) is a bit overkill though, so it might be better to only install `python3.5` and to get scons 3.1.2 via the `scons-local` upstream archive: https://scons.org/pages/download.html

PR #43, opened 2020-04-06
Factor out cloning and patching for godot-mono-builds

akien-mga:
This will also be used for iOS so it makes sense to do it all
together with the split out mono clone.

This also allows to pin the commit used more easily.

---

Yet untested, I'll do a follow up PR with iOS buildsystem changes for C# support and will use it to validate this first commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@
 *.zip
 *.xip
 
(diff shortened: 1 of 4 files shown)


akien-mga:
It seems to work as expected.

PR #44, opened 2020-04-06
Wasm: Fix mono patch application on Emscripten 1.38.47

akien-mga:
As a reminder, we pinned the Emscripten version to `1.38.47-upstream` as we had a regression with wasm generation in 1.38.48 and 1.39.x back in 17666f07fe.

This patch comes from Mono's `2019-10` branch (6.8.x): https://github.com/mono/mono/pull/17094
(Slightly amended to include a version bump in `wasm.mk` that should have been done by other intermediate patches.)

We might need a recent Emscripten 1.39.x in the `3.2` branch eventually (CC @Faless), so we might have to consider upgrading to a newer Mono branch with support for a more recent Emscripten.

### DIFF
--- a/Dockerfile.javascript
+++ b/Dockerfile.javascript
@@ -16,6 +16,7 @@ RUN if [ -z "${mono_version}" ]; then printf "\n\nArgument mono_version is manda
 RUN cp -a /root/files/${mono_version} /root && \
     cd /root/${mono_version} && \
     patch -p1 < /root/files/patches/mono-pr16636-wasm-bugfix-and-update.diff && \
+    patch -p1 < /root/files/patches/mono-pr17094-bump-emscripten-1.38.46.patch && \
     export MONO_SOURCE_ROOT=/root/${mono_version} && \
     export make="make -j" && \
     git clone --progress https://github.com/godotengine/godot-mono-builds /root/godot-mono-builds && \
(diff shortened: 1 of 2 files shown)


akien-mga:
For the record, this patch application failure should also have affected previous container builds used in Godot 3.2 and 3.2.1. I did not notice it until now because of godotengine/godot-mono-builds/issues/6.

I don't know what impact it has on our 3.2 and 3.2.1 WebAssembly templates for C#.

akien-mga:
> We might need a recent Emscripten 1.39.x in the `3.2` branch eventually (CC @Faless), so we might have to consider upgrading to a newer Mono branch with support for a more recent Emscripten.

I had a quick check at the supported Emscripten version in different Mono branches (I think we can assume that the hardcoded Emscripten version in Mono's `sdks/build/wasm.mk` is the version we *should* use, as Emscripten breaks compat in every other patch release):

- Mono `2019-08` (6.6.x, our current branch): [1.38.38](https://github.com/mono/mono/blob/2019-08/sdks/builds/wasm.mk) (hence the two patches we carry along)
- Mono `2019-10` (6.8.x, current upstream stable): [1.38.47](https://github.com/mono/mono/blob/2019-10/sdks/builds/wasm.mk)
- Mono `2019-12` (6.10.x, current upstream preview): [1.39.3](https://github.com/mono/mono/blob/2019-12/sdks/builds/wasm.mk)
- Mono `2020-02` (6.12.x, current upstream nightly): [1.39.7](https://github.com/mono/mono/blob/2020-02/sdks/builds/wasm.mk)

Latest emscripten today is 1.39.11.

So yeah, we should make sure that our Wasm port does not expect cutting edge Emscripten versions, unless we're willing to go with Mono Nightly for our stable releases...

We could consider an upgrade to Mono 6.10.x once it hits stable though, which gives us Emscripten 1.39.3 support without hacks.

akien-mga:
I confirm that the patches apply fine now when building a container against `mono-6.6.0.166`.

PR #53, opened 2020-07-08
podman is required.  also, "files" subdirectory here is incorrect.

arichnad:
podman is required.  "docker build -v" does not exist, this is only in podman.

"files" subdirectory here is incorrect.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,13 +1,10 @@
 #!/bin/bash
 set -e
 
-podman=`which podman`
-if [ -z $podman ]; then
-  podman=`which docker`
-fi
+podman=`which podman || true`
 
 if [ -z $podman ]; then
-  echo "Either podman or docker need to be in PATH for this script to work."
+  echo "podman needs to be in PATH for this script to work."
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


akien-mga:
> "files" subdirectory here is incorrect.

Why? It's this folder: https://github.com/godotengine/build-containers/tree/master/files
That's where the SDK archives should be put.

I've used it several times to build the Xcode image, it works fine.

arichnad:
> I've used it several times to build the Xcode image, it works fine.

@akien-mga I believe you have.  But please explain how this could possibly work:

`files_root=$(pwd)/files`
then
`${files_root}/files`

won't this be `$(pwd)/files/files` ?

I don't have a files/files, and I'm guessing you don't either?

akien-mga:
Ah I see, the change is correct indeed. This was a regression from 86739a8185a5, and apparently I didn't rebuild the Xcode container since then.

akien-mga:
Thanks!

arichnad:
Hey @akien-mga, sorry, I'm not sure who to contact about this:  How do I make this file: `aarch64-apple-darwin-mono-sgen.tar.xz`  It's fine if I need to use a mac, or something, just whatever notes you have on how to make this file, please let me know, thanks!

PR #58, opened 2020-10-20
Update to Mono 6.12.0.102, use Fedora 32

akien-mga:
- Base container: Fedora 32
- Mono uses version 6.12.0.102, MSBuild 16.6
- Linux builds use GCC 9
- Windows builds use Fedora 32's mingw-gcc (GCC 9)
- JavaScript uses Emscripten 1.39.9 (Mono 6.12 requires 1.39.7 but that version
  fails generating .wasm, so we backport a patch to get 1.39.9)
- Install SCons 4.0.1 from pip


akien-mga:
I used those containers successfully to make 3.2.4 beta 1, so merging.

PR #62, opened 2020-12-10
JavaScript: Install Emscripten 2.0.10 for standard builds with GDNative

akien-mga:
Mono 6.12 still forces us to stick to 1.39.9, so there won't be GDNative support
for it initially. But we can install 2.0.10 for standard builds.

The activation of the relevant toolchain will be done in the build scripts.

### DIFF
--- a/Dockerfile.javascript
+++ b/Dockerfile.javascript
@@ -6,13 +6,13 @@ ARG mono_version
 RUN if [ -z "${mono_version}" ]; then printf "\n\nArgument mono_version is mandatory!\n\n"; exit 1; fi && \
     dnf -y install --setopt=install_weak_deps=False \
       java-openjdk && \
-    git clone --progress https://github.com/emscripten-core/emsdk && \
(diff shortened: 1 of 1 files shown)


Faless:
I don't think this will work. Different versions are all installed in the same folder (`upstream/emscripten`).
See: https://github.com/Faless/build-containers/commit/280216a5410db3312a4bca617227e78267e32612

Faless:
And this for: https://github.com/Faless/godot-build-scripts/commit/ce72acc0973a44cbdbb6c05aaff070152620ac74
for the build scripts.

akien-mga:
Updated the PR. I still need to source Emscripten 1.39.9 for the Mono build in the container, but I kept the removal of the `~/.bashrc` sourcing (which didn't work anyway as it seems containers don't source `~/.bashrc` when called with a build script).

akien-mga:
Now seems to work fine :)

PR #64, opened 2021-01-05
Android: Update env variables and NDK preinstallation to match Godot buildsystem changes

m4gr3d:
Update configuration based on the Android buildsystem change in https://github.com/godotengine/godot/pull/44949.

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -11,10 +11,9 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mand
     unzip commandlinetools-linux-6609375_latest.zip && \
     rm commandlinetools-linux-6609375_latest.zip && \
     yes | tools/bin/sdkmanager --licenses && \
-    tools/bin/sdkmanager 'ndk;21.3.6528147' 'build-tools;30.0.1' 'platforms;android-30' 'cmake;3.10.2.4988404'
+    tools/bin/sdkmanager 'cmdline-tools;latest' 'build-tools;30.0.1' 'platforms;android-30' 'cmake;3.10.2.4988404'
 
 ENV ANDROID_SDK_ROOT=/root/sdk/
-ENV ANDROID_NDK_ROOT=/root/sdk/ndk/21.3.6528147/
 
 RUN cp -a /root/files/${mono_version} /root && \
     export MONO_SOURCE_ROOT=/root/${mono_version} && \
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -3,6 +3,9 @@ FROM godot-mono:${img_version}
 
 ARG mono_version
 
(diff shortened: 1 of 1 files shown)


akien-mga:
The issue with this change is that I'll have to redownload the NDK every single time I do Android builds, which kind of defeats the purpose of using a pre-built container with all dependencies installed.

Moreover, this will fail building as I need the NDK to build the Mono runtimes in that same container build script.

m4gr3d:
> The issue with this change is that I'll have to redownload the NDK every single time I do Android builds, which kind of defeats the purpose of using a pre-built container with all dependencies installed.
> 
> Moreover, this will fail building as I need the NDK to build the Mono runtimes in that same container build script.

I've updated the changes based on your feedback. We will need to ensure the `ANDROID_NDK_VERSION` is updated to match the project's `ndkVersion`. On the other hand, we can also make it an argument as for the `img_version` and `mono_version`.

Dockerfile.android
```diff
@@ -11,9 +14,9 @@ RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mand
     unzip commandlinetools-linux-6609375_latest.zip && \
     rm commandlinetools-linux-6609375_latest.zip && \
     yes | tools/bin/sdkmanager --licenses && \
-    tools/bin/sdkmanager 'cmdline-tools;latest' 'build-tools;30.0.1' 'platforms;android-30' 'cmake;3.10.2.4988404'
+    tools/bin/sdkmanager --sdk_root="${ANDROID_SDK_ROOT}" 'ndk;${ANDROID_NDK_VERSION}' 'cmdline-tools;latest' 'build-tools;30.0.1' 'platforms;android-30' 'cmake;3.10.2.4988404'
 
-ENV ANDROID_SDK_ROOT=/root/sdk/
+ENV ANDROID_NDK_ROOT=${ANDROID_SDK_ROOT}/ndk/${ANDROID_NDK_VERSION}
```
akien-mga, line 19: Can you move this up with the other env variables?
```
ENV ANDROID_SDK_ROOT=/root/sdk
ENV ANDROID_NDK_VERSION=21.3.6528147
ENV ANDROID_NDK_ROOT=${ANDROID_SDK_ROOT}/ndk/${ANDROID_NDK_VERSION}
```
(Also removed trailing `/` on SDK otherwise we get two `//` for NDK)

akien-mga:
> On the other hand, we can also make it an argument as for the `img_version` and `mono_version`.

It's fine hardcoded like done now IMO. It won't change often and it's not different from many other things hardcoded in build files like macOS and iOS SDK versions, etc.

Dockerfile.android
m4gr3d: Done!

akien-mga (approved):

akien-mga:
Thanks!

PR #66, opened 2021-01-18
Update Dockerfile.android

qjyqjyqjyqjy:
'号的错误 应该为”“


akien-mga:
Please try to be more explicit in your contributions, #65 did not say anything and this one is in Chinese, which I don't speak.

> '号的错误 应该为”“

Google Translate gives me:

> The error of the 'should be ""

So I think I now understand that to properly expand the `${ANDROID_NDK_VERSION}` variable it should indeed be within double quotes and not single quotes. That's correct.

akien-mga:
Thanks!

PR #70, opened 2021-02-10
Windows: Request net_4_x explicitly so that configure aborts if it can't be built

akien-mga:


akien-mga:
Well I wanted a confirmation that it does work before merging, as I couldn't reproduce the bug myself.

hpvb:
WE'LL DO IT LIVE

PR #75, opened 2021-03-03
Mention about SELinux and RAM requirements for building containers

### LINKED ISSUE #74: Build experience and issues on Fedora 33 and SELinux
#72 is merged and I've decided to give a try at compiling the engine this way. Also note that #12 has some advanced information as well (for those not familiar with `podman` or container technology in general). This is a summary of discussion at Godot Contributors chat.

I've managed to build images and managed to build something, but stumbled upon various errors in the process on a clear installation of Fedora 33 (which also made mono crash on Windows).

- [x] First thing first, do not attempt to use `sudo`, as `podman` is designed to work in rootless mode. Otherwise this may lead to permission errors once you build with user (for instance, failing to write to `logs` with `tee` may prevent other builds from being run at all).
- [x] Had to `dnf install patch` manually.
- [x] A lot of build steps resulted in permission errors with [SELinux enforced](https://www.phoronix.com/scan.php?page=news_item&px=Fedora-34-Drop-SELinux-Runtime). At first I added the `:z` tag to each `-v` path argument in `build.sh`, but it didn't resolve *all* permission errors, and even crashed in some cases (around building Windows).
     - Using `sudo setenforce 0` allows to circumvent this before building.
- [x] Also installed `dnf install wine` on the host to be sure, may be required.
- [x] Note sure if `mono` runtime should be installed on the host as well, as resulted in `mono: command not found` in some cases.
   - https://github.com/godotengine/build-containers/issues/74#issuecomment-789984899
- [x] Errors like `err:seh:install_bpf Unexpected status 0xffffffff, errno 1.`
   - seems to be innocent.
- [ ] Compilation error for macOS: https://github.com/godotengine/build-containers/issues/74#issuecomment-791549272.
- [ ] Build issues with UWP: https://github.com/godotengine/build-containers/issues/74#issuecomment-791615945.

Worth documenting that `export REGISTRY="localhost"` must be written in build script if building locally.

Feel free to close this at some point, I'd just like to document the limitations so I can refer to them in case I forget, and may be useful for other newbies like me.

Xrayez:
Closes #74.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,15 @@ you can comment out the corresponding lines from the script:
 
 ## Host OS preparation
 
+Before doing anything, make sure to disable Security-Enhanced Linux (SELinux) on
+hosts that have it enabled to prevent permission errors:
+
+    sudo setenforce 0
+
+Some build steps require enough of RAM to be available on your OS, so make sure
+you have a swap file/partition properly configured if your machine is low on RAM
+(aim for at least 8 GB swap size if you have 8 GB of RAM).
+
 ### Podman Fedora image
 
 To be extra-sure that you are building with the same base container image as the official


Xrayez:
The alternative to this is to check it at run-time, see changes here: https://github.com/goostengine/godot-build-scripts/commit/a36411904c5a5be369a381397ceebea77bcfcd37. This applies to both build container scripts and build scripts repositories.

akien-mga:
@hpvb Can you confirm if that was needed for you on the official buildsystem too?

Xrayez:
I've noticed that LTO builds fail using current build containers and scripts (the entire build fails miserably with `Terminated` message on LTO stage), so I wonder if the build requires even more RAM with LTO...

PR #88, opened 2021-07-29
Switch xcode-packer to use Fedora 33 (for the compat-openssl10-devel package)

dsnopek:
This image failed to build for me on `godot-fedora` because it appears that the compat-openssl10-devel package isn't available in Fedora 34. Switching to Fedora 33 allowed it to build for me!

I don't know if this is the best solution, but since this image is only used to build some files from the XCode.xip, it seems like it could be acceptable?

### DIFF
--- a/Dockerfile.xcode
+++ b/Dockerfile.xcode
@@ -1,5 +1,13 @@
 ARG img_version
-FROM godot-fedora:${img_version}
+#FROM godot-fedora:${img_version}
+FROM fedora:33
+
+WORKDIR /root
+
+RUN dnf -y upgrade --setopt=install_weak_deps=False && \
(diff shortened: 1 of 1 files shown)


akien-mga:
That can work, but I'd prefer to look into using latest openssl instead of having multiple base OSes (even though it doesn't matter much for this one which is just used to extract the SDK). I'll have a look.

dsnopek:
I tried using the 'openssl-devel' package (which would be openssl11), but the configure script for 'xar' didn't seem to like it. It's possible a newer version of 'xar' is needed to work with the newer openssl?

akien-mga:
I found a better fix :) #89.

dsnopek:
Awesome, thanks! Using the system package definitely seems way easier.

PR #89, opened 2021-07-30
Xcode: Fix xar compatibility with openssl 1.1

akien-mga:
Makes the script compatible with the current Fedora 34 base image.

Supersedes #88.


akien-mga:
Or maybe even simpler, we could likely use Fedora's existing `xar` package https://src.fedoraproject.org/rpms/xar :eyes: 

Do you remember why you didn't use it @hpvb? Too old at the time maybe? (It was since updated to track Apple upstream.)

akien-mga:
Superseded by #90.

PR #92, opened 2021-08-05
Windows: Fix GCC regression in Fedora 34's mingw-gcc

akien-mga:
Solves the crash reported in https://github.com/godotengine/godot/issues/50970 by patching Fedora 34's current mingw-gcc with the fix for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100402.

---

It adds 100 MiB to the repo so before merging, I'll wait a bit to see if Fedora pushes a fix soon, which could prevent me from having to include such big binaries in the Git history.

### DIFF
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -4,8 +4,14 @@ FROM godot-mono:${img_version}
 ARG mono_version
 
 RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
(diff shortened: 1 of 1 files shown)


akien-mga:
The update made it to Fedora 34: https://bodhi.fedoraproject.org/updates/FEDORA-2021-64d7f7f0ec

So this shouldn't be needed anymore, I'll just rebuild the Windows image ensuring that mingw-gcc-10.3.1-2.fc34 is pulled in.

PR #110, opened 2022-07-21
Rename OSX references to macOS

akien-mga:
Follow-up to https://github.com/godotengine/godot/pull/63225.

Draft for now because godot-build-scripts would need further adjustments. And it breaks compatibility for the images I currently use for both 3.5 and 4.0 builds, so I'm not in a hurry doing this change.


akien-mga:
Since Xcode still provides `MacOSX.sdk`, I don't think it's needed.

PR #112, opened 2022-08-20
Add a notice that the path to the scripts can't have spaces

### LINKED ISSUE #111: `build.sh` will fail if the path containing it has spaces
Every space separates the path into new args.

I believe many people looking for a reliable way to build Godot binaries will stop by this repo and may not know this.

I'm thinking about adding a notice about the path.

DeeJayLSP:
Closes #111 by adding a notice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,9 @@ tried we ran into performance issues).
 
 ## Usage
 
+**Note:** Make sure the path containing the scripts doesn't have spaces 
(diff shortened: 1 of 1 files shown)


Calinou:
Shouldn't we try to fix the original issue by quoting paths instead?

umarcor:
I believe this was superseeded by #121.

PR #113, opened 2022-08-29
.NET 6: Remove all Mono build code, install .NET SDK 6.0

akien-mga:
Draft for now as this is still being validated.

To use with build scripts from https://github.com/godotengine/godot-build-scripts/pull/64.


akien-mga:
I've tested these containers successfully for a non-dotnet build (there are still some outstanding issues to solve for dotnet linking due to the use of `libnethost.a`, but generally it seems to compile fine with Fedora's `dotnet-sdk-6.0`).

So I'll merge now since I'm going to use these containers for 4.0 alpha 15.

PR #119, opened 2023-04-08
Support executing `build.sh` from anywhere

umarcor:
Instead of requiring the script to be executed from the root of the repo due to `$(pwd)/files` and `files/`, use `$(dirname "$0"; pwd)/files` and `$files_root`.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -26,7 +26,7 @@ fi
 godot_branch=$1
 base_distro=$2
 img_version=$godot_branch-$base_distro
-files_root=$(pwd)/files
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

umarcor:
@akien-mga can we please have this merged?

Calinou (approved):
Thanks!

umarcor:
@akien-mga that's interesting. I believe I tried it on Fedora (37), Ubuntu (22) and Debian (Bullseye). I would expect such issues with shells such as ash or dash, but not with bash. I will try again on Fedora, if only for curiosity.

PR #120, opened 2023-04-08
build: rm base_distro

umarcor:
On the one hand, the base distro is hardcoded in the Dockerfiles. On the other hand, there are not multiple variants of the images using different base distros. Moreover, these images are expected to used along with scripts from godot-build-scripts. Therefore, the base distro is not very relevant in the name. In any case, it should not be a CLI argument which is independent from the hardcoded value.


akien-mga:
I'm not fond of that change. The base distro defines a lot of the software we have available in the series of images, and while it's an approximation, it can usually be tied to a major update of these images.

For example now I'm considering doing an update of all images based on Fedora 38 with newer toolchains. I still want to keep my existing images for comparison, and if they would both be named `4.x`, that's not much to go on.

We could introduce an arbitrary version number, but so far relying on the base distro was good enough for the `4.x` branch (and for `3.x` it's base distro + Mono version, as that one is updated more frequently).

PR #122, opened 2023-04-09
build: skip confirm if non-interactive

umarcor:
Skip confirmation if the build script is run in a non-interactive shell (for instance, in CI).


akien-mga:
Thanks!

PR #121, opened 2023-04-08
build: make podman_build a function

umarcor:
This PR is based on #119.
It should fix #111 and superseed #112.

PR #123, opened 2023-04-09
build: support specifying subset of platforms

umarcor:
This PR is based on #121.

Instead of building all the images when `./build.sh` is executed, this PR allows specifying a subset through a comma-separated list.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -43,36 +43,45 @@ done
 
 mkdir -p logs
 
-# You can add --no-cache  as an option to podman_build below to rebuild all containers from scratch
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -4,8 +4,10 @@ basedir=$(cd $(dirname "$0"); pwd)
 
 source $basedir/setup.sh
 
+platforms="linux,windows,web,android,osx,ios,msvc"
```
akien-mga, line 7: The `msvc` platform doesn't work currently, so it shouldn't be built by default. That's why it's disabled in `main`.

Virus-Axel:
This is a pretty neat feature to have when building images from another SCons file. What is the status of this? How about removing the msvc from default platforms?

PR #128, opened 2023-09-05
Update to Fedora 39, update all toolchains

akien-mga:
Update to Fedora 39, update all toolchains

- Base image: Fedora 39
- SCons: 4.5.2
- Linux: GCC 10.2.0 built against glibc 2.19, binutils 2.35.1
- Windows: MinGW 11.0.0, GCC 13.2.1, binutils 2.40
- Web: Emscripten 3.1.39
- Android: Android NDK 23.2.8568313, build-tools 33.0.2, platform android-33, CMake 3.22.1, JDK 17
- macOS: Xcode 15.0 with Apple Clang (LLVM 16.0.0), MacOSX SDK 14.0
- iOS: Xcode 15.0 with Apple Clang (LLVM 16.0.0), iPhoneOS SDK 17.0

Switches macOS/iOS builds to use Apple Clang, to maximize compatibility with upstream Xcode. ~This enables us to build the arm64 simulator for iOS.~ (Still doesn't work.)

Also add gettext which is needed by Godot build scripts to compile PO files to MO.

---

Draft for now as I haven't tested building Godot with those yet. I expect at least for Android we need some adjustments in upstream build scripts to use those toolchains (CC @m4gr3d).

### DIFF
--- a/Dockerfile.ios
+++ b/Dockerfile.ios
@@ -1,30 +1,31 @@
 ARG img_version
 FROM godot-osx:${img_version}
 
-ENV IOS_SDK=16.1
+ENV IOS_SDK=17.0
 
 RUN dnf -y install --setopt=install_weak_deps=False \
       automake autoconf gcc gcc-c++ gcc-objc gcc-objc++ cmake libicu-devel libtool libxml2-devel openssl-devel perl python yasm && \
-    git clone --progress https://github.com/tpoechtrager/cctools-port.git && \
+    git clone --progress https://github.com/tpoechtrager/cctools-port && \
     cd /root/cctools-port && \
-    git checkout 11c93763d7e7ce7305163341d08052374e4712de && \
+    git checkout 437ced391dbf14dce86f977ca050a750d5682f39 && \
+    curl -LO https://github.com/tpoechtrager/cctools-port/pull/136.patch && \
+    git apply 136.patch && \
     # arm64 device
     usage_examples/ios_toolchain/build.sh /root/files/iPhoneOS${IOS_SDK}.sdk.tar.xz arm64 && \
     mkdir -p /root/ioscross/arm64 && \
     mv usage_examples/ios_toolchain/target/* /root/ioscross/arm64 && \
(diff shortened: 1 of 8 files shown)


Dockerfile.ios
```diff
@@ -1,30 +1,31 @@
...
     mkdir -p /root/ioscross/arm64 && \
     mv usage_examples/ios_toolchain/target/* /root/ioscross/arm64 && \
     mkdir /root/ioscross/arm64/usr && \
     ln -s /root/ioscross/arm64/bin /root/ioscross/arm64/usr/bin && \
+    # Prepare for simulator builds
+    sed -i '/WRAPPER_SDKDIR/s/iPhoneOS/iPhoneSimulator/' usage_examples/ios_toolchain/build.sh && \
     # arm64 simulator
-    # Disabled for now as it doesn't work with cctools-port and current LLVM:
-    # Cf. https://github.com/tpoechtrager/cctools-port/issues/102
-    #sed -i 's/miphoneos-version-min/mios-simulator-version-min/g' usage_examples/ios_toolchain/wrapper.c && \
```
akien-mga, line 20: For future reference, we discussed removing this with @bruvzg on RC:
```
<Akien> BTW, were our simulator builds a bit wonky due to this file being commented out, even though it seems relevant for both arm64_sim and x86_64_sim? https://github.com/godotengine/build-containers/blob/main/Dockerfile.ios#L20
<bruvzg> It seems to be for signing, which is irrelevant since we build static libs only. x86_64 simulator build were functional.
<bruvzg> And the rest of the args it's setting should be already set by Scons.
<Akien> Should I remove that line then to simplify the script?
<bruvzg> Probably, it was never used.
```

akien-mga:
I downgraded Emscripten to 1.3.39 to solve https://github.com/godotengine/godot/issues/82865.

Also kept Android NDK to r23 and build-tools/platform to API 33, matching the upstream gradle config. We should look into targeting API 34 before it becomes a requirement next Summer though.

PR #134, opened 2023-12-11
Update Linux SDK to 2023-12-11 release, fixes arm64 Rasperry Pi compatibility

akien-mga:
The only change is https://github.com/godotengine/buildroot/commit/e6c2704eb22f0ea2a5dbe3d3084cec414ae540c8:

> Build with 16K page sizes in mind on aarch64
>
> By default binutils on buildroot aligns sections to 4kb boundaries, but
> there are several 16k only linux distros out there now that default to
> 16kb pages.
>
> With this change the elf files will have their sections aligned to 16kb
> boundaries which should ensure binaries produced with the toolchain work
> on most aarch64 linuxes.


akien-mga:
Tested successfully for 4.2.1-stable.

PR #135, opened 2024-01-16
[3.x] Update to Fedora 39, newer toolchains, add Linux arm64

akien-mga:
Syncs with the config of the current `main` branch (as used for 4.2+) as far as possible (backports the last couple of years of PRs, minus the removal of some platforms and of the Mono builds).

Notably backports #128 and #131, giving us these toolchains:

- Base image: Fedora 39
- Mono version: 6.12.0.198
- SCons: 4.5.2
- Linux: GCC 13.2.0 built against glibc 2.28, binutils 2.40, from our own [Linux SDK](https://github.com/godotengine/buildroot)
- Windows: MinGW 11.0.0, GCC 13.2.1, binutils 2.40
- HTML5: Emscripten 3.1.39 (standard builds), Emscripten 1.39.9 (Mono builds)
- Android: Android NDK 23.2.8568313, build-tools 33.0.2, platform android-33, CMake 3.22.1, JDK 11
- macOS: Xcode 15.0 with Apple Clang (LLVM 16.0.0), MacOSX SDK 14.0
- iOS: Xcode 15.0 with Apple Clang (LLVM 16.0.0), iPhoneOS SDK 17.0
- UWP: Unchanged, current Dockerfile still unsupported

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -7,16 +7,15 @@ ENV ANDROID_SDK_ROOT=/root/sdk
 ENV ANDROID_NDK_VERSION=23.2.8568313
 ENV ANDROID_NDK_ROOT=${ANDROID_SDK_ROOT}/ndk/${ANDROID_NDK_VERSION}
 
-RUN if [ -z "${mono_version}" ]; then echo -e "\n\nargument mono-version is mandatory!\n\n"; exit 1; fi && \
-    dnf -y install --setopt=install_weak_deps=False \
-      java-11-openjdk-devel ncurses-compat-libs && \
+RUN dnf -y install --setopt=install_weak_deps=False \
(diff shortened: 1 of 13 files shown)


akien-mga:
For the record, there are some newer versions of some of the toolchains listed above. This PR syncs with #128 which was mostly up-to-date as of Sep 2023, and since we've tested those versions well by now, I'm using the same for `3.x`.

As for Mono, 6.12.0.198 is the last version we built on https://github.com/godotengine/godot-mono-builds/releases/tag/release-4912f62 so I just reused that.
Technically the Mono website lists 6.12.0.199 as the stable version (the only change seems to be updating the nuget artifact they're using, I'm not sure it affects us). And there's a few commits on top in the Preview version. We might want to upgrade further eventually, but for now this should be perfectly fine.

PR #141, opened 2024-06-19
Update various toolchains for 4.3

akien-mga:
- Fedora 40
- SCons 4.7.0
- MinGW 11.0.1 with GCC 14.1.1 and binutils 2.41
- Xcode 15.5 (macOS 14.5, iOS 17.5)
- Emscripten 3.1.61

### DIFF
--- a/Dockerfile.web
+++ b/Dockerfile.web
@@ -1,7 +1,7 @@
 ARG img_version
 FROM godot-fedora:${img_version}
 
-ENV EMSCRIPTEN_VERSION=3.1.53
+ENV EMSCRIPTEN_VERSION=3.1.61
 
 RUN git clone --branch ${EMSCRIPTEN_VERSION} --progress https://github.com/emscripten-core/emsdk && \
     emsdk/emsdk install ${EMSCRIPTEN_VERSION} && \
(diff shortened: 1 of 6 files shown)


Dockerfile.web
```diff
@@ -1,7 +1,7 @@
 ARG img_version
 FROM godot-fedora:${img_version}
 
-ENV EMSCRIPTEN_VERSION=3.1.53
+ENV EMSCRIPTEN_VERSION=3.1.61
```
akien-mga, line 4: We'll want to bump to 3.1.62 as soon as it's released, and maybe aim to release 4.3 with that if tests are conclusive?

CC @adamscott @Faless

akien-mga:
I've used these containers for 4.3-beta2.

Dockerfile.web
Faless: Yeah, I think 62 will most likely be released before beta3, I don't think it's worth rolling it back (39 isn't perfect either).

Worst case scenario we can apply the needed patch after installing 61.

We could also try politely asking upstream for a new release next week :sweat_smile:

akien-mga: > Worst case scenario we can apply the needed patch after installing 61.

I did that for now, I don't want to rush upstream needlessly. I complain enough when they make a release that breaks our build so if they need time to finalize 3.1.62, I'll give it to them :D 

https://github.com/godotengine/build-containers/pull/142

PR #147, opened 2024-08-21
Add `.DS_Store` to .gitignore

adamscott:
Does what the tin says.


Calinou:
I suggest adding that file to your [global gitignore](http://stackoverflow.com/questions/7335420/), but this makes sense to add for consistency with the main Godot repository.

PR #149, opened 2024-08-22
Use extract_xcode.py instead of compiling pbzx

adamscott:
This PR adds extract_xcode.py to do what it says.

Capture d’écran, le 2024-08-22 à 09 09 17

For an unknown reason, as shown above, `xar` of the VM refuses to extract the Xcode_15.4.xip file on my MacBook Pro. (for the record, the SHA256 sum of the file is alright). But with [extract_xcode.py](https://github.com/bitcoin-core/apple-sdk-tools/blob/master/extract_xcode.py), it works OK. As extract_xcode.py contains the pbzx algorithm, no need to compile the C project.

I suggest we replace the existing algorithm for compatibility reasons.


zynga-jpetersen:
While this branch did get me passed the xar extract, which is great, I still wasnt able to compile osx, getting the following errors:

[osx.log](https://github.com/user-attachments/files/16966665/osx.log)

-- Performing Test HAVE_CXX_FLAG_WD654
-- Performing Test HAVE_CXX_FLAG_WD654 - Failed
-- Performing Test HAVE_CXX_FLAG_WTHREAD_SAFETY
-- Performing Test HAVE_CXX_FLAG_WTHREAD_SAFETY - Failed
-- Performing Test HAVE_CXX_FLAG_COVERAGE
-- Performing Test HAVE_CXX_FLAG_COVERAGE - Success
-- Performing Test HAVE_GNU_POSIX_REGEX
-- Performing Test HAVE_GNU_POSIX_REGEX
-- Performing Test HAVE_GNU_POSIX_REGEX -- failed to compile

I'm currently confirming what the differences are between this and a correctly built version (which I'm able to do on my windows machine under wsl2),but I have to regenerate the container to get the logs to compare the build.

PR #151, opened 2025-05-16
Update various toolchains for 4.5

akien-mga:
- Fedora 42
- SCons 4.9.1
- Windows: MinGW 12.0.0 with GCC 14.2.1 and binutils 2.43.1
- Android: NDK r28b, API 25, Java 21
  * CC @m4gr3d - Depends on https://github.com/godotengine/godot/pull/106152, https://github.com/godotengine/godot/pull/106358, https://github.com/godotengine/godot/pull/106466
- Web: Emscripten 4.0.8
  * CC @adamscott - I'm not sure we support it yet, and so far we've still built with 3.1.64, but it's probably time to start considering which version we want to use for 4.5.
- Apple: Xcode 16.3 (macOS 15.4, iOS 18.4)
  * Will open a follow-up that starts working on visionOS support.
- Linux: No change, still GCC 13.2.0
  * CC @hp - Our buildroot fork is based on `2023.08.x`, and the latest branch upstream is https://github.com/buildroot/buildroot/tree/2025.02.x. Might be a good time now to update and go from GCC 13.2.0 to 14.2.0.

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -2,17 +2,17 @@ ARG img_version
 FROM godot-fedora:${img_version}
 
 ENV ANDROID_SDK_ROOT=/root/sdk
-ENV ANDROID_NDK_VERSION=23.2.8568313
+ENV ANDROID_NDK_VERSION=28.1.13356709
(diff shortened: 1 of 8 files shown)


m4gr3d:
The Android part looks good.

Did you mean API 24 in the description?

PR #71, opened 2021-02-12
[WIP] tvOS build support

naithar:
Adds a container which can be used to build tvOS binaries.
Depends on godotengine/godot#45829 and https://github.com/godotengine/godot-build-scripts/pull/21

PR #152, opened 2025-05-16
WIP: visionOS build container

akien-mga:
- Depends on and includes #151.

Initial WIP on adding support for compiling for visionOS using cctools-port, like we do for iOS.

For now this only adds support for extracting the `XROS` and `XRSimulator` SDKs from the Xcode xip, so that's not doing much.

To (attempt to) make a build image that can target visionOS, we should copy `Dockerfile.ios` to a new `Dockerfle.visionos` and see how to change it to target this related platform.

The work done by @naithar for tvOS a few years ago should be helpful:
- #71

Notably we might indeed need to write a custom wrapper, or patch cctools-port's current one in https://github.com/tpoechtrager/cctools-port/tree/1024.3-ld64-955.13/usage_examples/ios_toolchain to make it target visionOS instead of iOS.

I note that there's at least one user of cctools-port who tried to do that and had some issue:
- https://github.com/tpoechtrager/cctools-port/issues/149

I don't know if we'd run into the same, and if it's something we can help fix upstream.

CC @bruvzg @stuartcarnie @BastiaanOlij

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -2,17 +2,17 @@ ARG img_version
 FROM godot-fedora:${img_version}
 
 ENV ANDROID_SDK_ROOT=/root/sdk
-ENV ANDROID_NDK_VERSION=23.2.8568313
+ENV ANDROID_NDK_VERSION=28.1.13356709
 ENV ANDROID_NDK_ROOT=${ANDROID_SDK_ROOT}/ndk/${ANDROID_NDK_VERSION}
 
 RUN dnf -y install --setopt=install_weak_deps=False \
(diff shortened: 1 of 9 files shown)


bruvzg:
> Notably we might indeed need to write a custom wrapper, or patch cctools-port's current one in

Probably, since it's adding `-miphoneos-version-min` and it should be replaced with `-mtargetos=xros2.0`. Platforms should be similar, so at worst case, I would expect patching `LC_BUILD_VERSION` (like we did with iOS simulator).

akien-mga:
Superseded by #154.

PR #153, opened 2025-05-16
Update osx env variable to 16.2

svelezw4games:
Description:
THis PR updates the env variable to the latest xcode sdk version 16.2.

Reason:
If we build this container, with the latest package, it could result in an error since the file name is dynamically created with the ENV variable

```
extracting MacOSX15.2.sdk.tar.xz ...
xz: /root/osxcross/tarballs/MacOSX15.2.sdk.tar.xz: No such file or directory
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
```

### DIFF
--- a/Dockerfile.osx
+++ b/Dockerfile.osx
@@ -1,7 +1,7 @@
 ARG img_version
 FROM godot-fedora:${img_version}
 
-ENV OSX_SDK=15.2
+ENV OSX_SDK=16.2
 
 RUN dnf -y install --setopt=install_weak_deps=False \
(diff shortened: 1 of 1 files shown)


akien-mga:
Hi @svelezw4games :)

That's incorrect, 15.2 *is* the version of the MacOSX SDK in Xcode 16.2.
SDKs and Xcode are versioned separately, see this table for each version:
https://en.wikipedia.org/wiki/Xcode#Xcode_15.0_-_16.x_(since_visionOS_support)

See the versions in `build.sh` for when `Dockerfile.xcode` is built to extract the various SDKs:
https://github.com/godotengine/build-containers/blob/main/build.sh#L61-L81

PR #158, opened 2025-10-27
[3.x] Update to Fedora 43 for MinGW-GCC 15.2.1

### LINKED ISSUE #157: [3.x] MinGW-GCC 14.2.1 from Fedora 42 causes Windows editor crashes (AssetLib, installing export templates)
Follow-up to:
- #156 
- https://github.com/godotengine/godot/issues/111979
- https://github.com/godotengine/godot/issues/112021

It turns out that the upgrade from Fedora 39 to Fedora 42 (thus updating MinGW-GCC) introduced a regression for Windows builds on the 3.6.2-stable release.

We then rebuilt 3.6.2-stable Windows binaries with the previous Fedora 39-based container to work it around. But this needs to be investigated so we can go back to Fedora 42 (we also have an inconsistent state for 3.6.2-stable and potential future builds as we need some of the changes from the Fedora 42-based images to build Android and Apple platforms at least).

The issue happens only with builds made with LTO (`production=yes`). This isn't the first time we have such issue and in the past it could be traced back to either GCC or binutils bugs specifically affecting MinGW/Win32.

Potential options:
- Test if MinGW-GCC 15.2.1 in Fedora 43 works better.
- Figure out the MinGW-GCC bug and patch it.
- Revert the Windows image to Fedora 39, or install its MinGW-GCC version directly on Fedora 42 (or any earlier version that works, could be Fedora 40 or 41 if it's tested).
- Switch to llvm-mingw if it's tested to work better.

CC @lawnjelly FYI. This needs to be resolved one way or another for 3.7-dev1 and 3.6.3-stable.

akien-mga:
Fixes #157, seems like the MinGW stack in F42 was broken.

Still needs testing to validate the changes.

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -4,7 +4,7 @@ FROM godot-mono:${img_version}
 ARG mono_version
 
 ENV ANDROID_SDK_ROOT=/root/sdk
-ENV ANDROID_NDK_VERSION=23.2.8568313
+ENV ANDROID_NDK_VERSION=28.1.13356709
 ENV ANDROID_NDK_ROOT=${ANDROID_SDK_ROOT}/ndk/${ANDROID_NDK_VERSION}
 
 RUN dnf -y install adoptium-temurin-java-repository && \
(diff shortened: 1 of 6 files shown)


akien-mga:
Tested successfully on a potential build for 3.7-dev1, and the Windows editor seems to work without the bugs listed in #157.

PR #159, opened 2025-11-22
Update various toolchains for 4.6

akien-mga:
- Fedora 43
- SCons 4.10.1
- MinGW 13.0.0 with GCC 15.2.1 and binutils 2.45
- Emscripten 4.0.20
- Xcode 26.1.1


akien-mga:
~We've been using this in production for a few weeks (4.6 dev6 and beta1 at least),~ I just forgot to merge this.

*Edit:* Apparently no, see below.

bruvzg:
> MinGW 13.0.0 with GCC 15.2.1 and binutils 2.45

This will likely require changing this two lines:
https://github.com/godotengine/godot-build-scripts/blob/5261d2c468a9ad3dcf643ba2174105402a3a8b4b/build.sh#L186-L187

```diff
-  curl -L -o windows_x86_64.zip $base_url-x86_64-gcc-release.zip
-  curl -L -o windows_x86_32.zip $base_url-x86_32-gcc-release.zip
+  curl -L -o windows_x86_64.zip $base_url-x86_64-gcc-13-release.zip
+  curl -L -o windows_x86_32.zip $base_url-x86_32-gcc-13-release.zip
```

bruvzg:
> We've been using this in production for a few weeks (4.6 dev6 and beta1 at least), I just forgot to merge this.

Or if it's working as is, it might be only backward incompatible and lib built with MinGW 12 will work with MinGW 13. Probably better to change it anyway to match runtime version.

akien-mga:
Yeah it probably makes sense to change those, I missed that we had two sets of binaries for ANGLE.

akien-mga:
Updated with https://github.com/godotengine/godot-build-scripts/commit/9fb0561da557d9fffacbea4da5a659898c07c90e.

akien-mga:
> We've been using this in production for a few weeks (4.6 dev6 and beta1 at least), I just forgot to merge this.

It turns out that we had *not* been using this in production yet, I was mistaken. 4.6 dev6 and beta1 were still built with the previous Fedora 42-based set of images.

So the Windows builds we made today for beta2 failed due to the ANGLE version. I restarted a new build, this time properly switching to the Fedora 43 images I intend to use for stable (unless we find critical regressions).

PR #161, opened 2026-03-17
[Linux] Fix separate debug symbols for arm

AThousandShips:
Buildroot is missing aliases for `objcopy` and `strip` meaning that it will fall back to using the default, which doesn't have support for arm binaries.

This creates these aliases for the respective platforms (including x86_64 and 32 even though they do work with the default objcopy/strip, have not compared the results between the two but as we create aliases for other build tools I think it's best to do so here too to ensure we use buildroot consistently)

Will make an issue to track improving this as part of buildroot as well


akien-mga (approved):

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

PR #162, opened 2026-04-11
Apple: visionOS and tvOS builds now supported; update to 26.4 SDKs

stuartcarnie:
Changes in this PR:

- Updates Apple SDKs to 26.4.
- Installs the Swift 6.3 toolchain via swiftly in `Dockerfile.osx` so visionOS Swift sources can be cross-compiled from Linux.
- Replaces the local `osxcross-fix-visionos.patch` with the upstream fix from [swiftlang/llvm-project#12174](https://github.com/swiftlang/llvm-project/pull/12174), adapted for the clang 19.1.5 tree in osxcross (`AvailabilityMergeKind::None` → `AMK_None`).
- Fixes visionOS SDK / Swift toolchain incompatibilities at image build time via a new `files/appleembedded/patch-visionos-sdk.sh`, invoked from `Dockerfile.appleembedded`, which rewrites the visionOS SDK in place:
  - Adds `visionos` to `API_UNAVAILABLE(...)` header lists wherever `ios` is already listed, working around [swiftlang/llvm-project#11866](https://github.com/swiftlang/llvm-project/pull/11866) not yet being merged into `swift/release/6.3`.
  - Rewrites `// swift-compiler-version:` in every `.swiftinterface` so the OSS Swift 6.3 release compiler accepts Apple's interfaces (the marketing version matches but the swiftlang stamps differ: `swiftlang-6.3.0.123.x` vs `swift-6.3-RELEASE`).
  - Guts `@inlinable` bodies in `RealityFoundation.swiftinterface` that reference C++ types from `simd`/`AVFAudio` (library-evolution errors under `-cxx-interoperability-mode=default`), and swaps the `Audio.GeneratorRenderHandler` typealias to a function type so call sites using `@escaping` still parse.
  - Replaces `CTTextAlignment`/`CTLineBreakMode` default-argument expressions in `RealityKit.swiftinterface` with raw-value initializers, since the OSS compiler strips C enum prefixes less aggressively than Apple's and the bare `.left` / `.byTruncatingTail` forms don't resolve.

All SDK patches are idempotent, visionOS-only (they touch `XROS.sdk`), and have comments pointing at the relevant upstream issue so each can be removed once the corresponding fix lands.

### DIFF
--- a/Dockerfile.osx
+++ b/Dockerfile.osx
@@ -5,15 +5,16 @@ FROM godot-fedora:${img_version}
 #
 # sudo dnf install gnupg2
 
-ENV XCODE_SDKV=26.1.1
-ENV APPLE_SDKV=26.1
+ENV XCODE_SDKV=26.4
+ENV APPLE_SDKV=26.4
 
 RUN dnf -y install --setopt=install_weak_deps=False \
       automake autoconf bzip2-devel cmake gawk gcc gcc-c++ libdispatch libicu-devel libtool \
       libxml2-devel openssl-devel uuid-devel yasm gpg && \
     git clone --progress https://github.com/tpoechtrager/osxcross && \
     cd /root/osxcross && \
-    git checkout 121ce150c7857a9474dfff8a8e431482806b3e1b && \
+    # latest commit from main branch
+    git checkout e6ab3fa7423f9235ce9ed6381d6d3af191b46b59 && \
     # Patch to fix visionOS support.
     # See: https://github.com/llvm/llvm-project/issues/142502
     patch -p1 < /root/files/patches/osxcross-fix-visionos.patch && \
@@ -31,7 +32,7 @@ RUN dnf -y install --setopt=install_weak_deps=False \
 ENV OSXCROSS_ROOT=/root/osxcross
 ENV PATH="/root/osxcross/target/bin:${PATH}"
 
-# Install Swift 6.2 toolchain
+# Install latest Swift toolchain
 
 RUN mkdir -p /root/Xcode.app/Contents/Developer
 RUN cd /root/Xcode.app/Contents/Developer && tar xf /root/files/Xcode-Developer${XCODE_SDKV}.tar.xz --strip-components=1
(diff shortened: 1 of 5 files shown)


Dockerfile.osx
akien-mga, line 27: For Xcode 26.4, we should use Clang 21.1.6: https://en.wikipedia.org/wiki/Xcode#Xcode_26.0_(since_version_number_change)_2

Dockerfile.osx
```diff
@@ -5,15 +5,16 @@ FROM godot-fedora:${img_version}
...
+ENV XCODE_SDKV=26.4
+ENV APPLE_SDKV=26.4
 
 RUN dnf -y install --setopt=install_weak_deps=False \
       automake autoconf bzip2-devel cmake gawk gcc gcc-c++ libdispatch libicu-devel libtool \
       libxml2-devel openssl-devel uuid-devel yasm gpg && \
     git clone --progress https://github.com/tpoechtrager/osxcross && \
     cd /root/osxcross && \
-    git checkout 121ce150c7857a9474dfff8a8e431482806b3e1b && \
+    # latest commit from main branch
```
akien-mga, line 16:
```suggestion
    # Latest commit from main branch.
```

Dockerfile.osx
akien-mga, line 42: We should pin the version of Swift we're installing, to make sure it matches the Xcode SDK we're compiling for, and have reproducible container builds.

Seems like this can be done with something like:
```
./swiftly init --platform ubi9 --quiet-shell-followup --skip-install -y
~/.local/share/swiftly/bin/swiftly install 6.3.0
```

And as suggested in https://github.com/godotengine/godot-build-scripts/pull/156#discussion_r3075582174 I'd add a `ENV SWIFT_VERSION=6.3.0` that can be used here and in the build scripts.

stuartcarnie: I'll test that to make sure everything compiles with `21.1.6`

stuartcarnie:
Closed in favour of the New Improved™ 

- #163

PR #163, opened 2026-04-15
Apple: simplified build system for all platforms

stuartcarnie:
> [!IMPORTANT]
>
> We should use the 6.4.x-snapshot-2025-06-11 for all builds, as those are still official releases from Apple

* single build container, `godot-apple:<image version>` for all Apple platforms
* Gated by `APPLE_LLVM_CROSS=1`
* visionOS+tvOS support
* SDK 26.5
* Swift 6.3.2 (via ENV variable)
* Swift 6.4.x-snapshot
  * via `--build-arg swift_version=6.4.x-snapshot-2026-06-11`
* uses clang from Swift install
* uses llvm-lipo to create Universal binaries
* uses [xtool-darwin-tools-llvm](https://github.com/xtool-org/darwin-tools-linux-llvm) release for lld
* no longer requires [custom MoltenVK build](https://github.com/godotengine/moltenvk-osxcross) for `-fno-objc-msgsend-selector-stubs -Wno-unused-command-line-argument`
* no longer depends on osxcross

As a win, building a single container for all Apple platforms is significantly faster 🚀 

Depends on:

- https://github.com/godotengine/godot-build-scripts/pull/159
- https://github.com/godotengine/godot/pull/121255


raulsntos:
I have tested this (and the related PR on `godot-build-scripts`) to build the Godot .NET prerelease.

Before this PR, I was having issues building the iOS export templates, but it seems to work now although I haven't actually tested using the templates. This could just be a environment issue on my side, not related to this change at all.

I did get some warnings in the logs. I'm not sure if this is my setup's fault so I'll post the full logs here just in case. The warnings mention something about not finding clang intrinsic headers and that it should be reported to the OSXCross project.

[logs.zip](https://github.com/user-attachments/files/28211411/logs.zip)

I also had to comment out the part about notarization because it seems the `sign.sh` script is missing in `godot-build-scripts`.

stuartcarnie:
@raulsntos those warnings are currently expected, but don't affect the compiled output.

Thanks for testing!

Dockerfile.apple
```diff
@@ -62,7 +62,7 @@
 #
 # `osxcross-add-sdk-26.patch` adds `darwin25.x` target
 # mappings to osxcross's `build.sh` so `SDK_VERSION=26.5 ./build.sh`
-# recognizes the SDK.
+# recognizes the SDK. (Xcode 26.6 bundles the macOS 26.5 SDK.)
```
akien-mga: That's so confusing. I thought they had finally agreed to streamline everything around the same numbers with Xcode 26...

So we have Xcode 26.6 providing MacOSX SDK 26.5 which maps to darwin 25.6 binaries. Not confusing at all 😅

stuartcarnie: Yah, I even see it on macOS and in the build when it resolves the SDKs it says 26.5. I actually think iOS is 26.5.1, which is the only one that was rev'd

files/patches/osxcross-add-sdk-26.patch
```diff
@@ -0,0 +1,16 @@
...
+--- a/build.sh	(revision e6ab3fa7423f9235ce9ed6381d6d3af191b46b59)
++++ b/build.sh	(date 1779656427653)
+@@ -68,5 +68,9 @@
+   26|26.0*) TARGET=darwin25;   SUPPORTED_ARCHS="arm64 arm64e x86_64 x86_64h"; NEED_TAPI_SUPPORT=1; OSX_VERSION_MIN_INT=10.13 ;;
+   26.1*) TARGET=darwin25.1;   SUPPORTED_ARCHS="arm64 arm64e x86_64 x86_64h"; NEED_TAPI_SUPPORT=1; OSX_VERSION_MIN_INT=10.13 ;;
+   26.2*) TARGET=darwin25.2;   SUPPORTED_ARCHS="arm64 arm64e x86_64 x86_64h"; NEED_TAPI_SUPPORT=1; OSX_VERSION_MIN_INT=10.13 ;;
++  26.3*) TARGET=darwin25.3;   SUPPORTED_ARCHS="arm64 arm64e x86_64 x86_64h"; NEED_TAPI_SUPPORT=1; OSX_VERSION_MIN_INT=10.13 ;;
++  26.4*) TARGET=darwin25.4;   SUPPORTED_ARCHS="arm64 arm64e x86_64 x86_64h"; NEED_TAPI_SUPPORT=1; OSX_VERSION_MIN_INT=10.13 ;;
++  26.5*) TARGET=darwin25.5;   SUPPORTED_ARCHS="arm64 arm64e x86_64 x86_64h"; NEED_TAPI_SUPPORT=1; OSX_VERSION_MIN_INT=10.13 ;;
++  26.6*) TARGET=darwin25.6;   SUPPORTED_ARCHS="arm64 arm64e x86_64 x86_64h"; NEED_TAPI_SUPPORT=1; OSX_VERSION_MIN_INT=10.13 ;;
```
akien-mga: Isn't it still `darwin25.5` like for Xcode 26.6?
As we discussed below it seems 26.6 is actually a misnomer for 26.5.1 and only for iOS.

stuartcarnie: Correct – I'll remove the 25.6 patch, given it doesn't yet exist.

bruvzg (approved):

akien-mga (approved):
Tested successfully, it seems to work great!

I haven't reviewed everything in depth in the shell scripts and `Dockerfile.apple`, but it seems relatively straightforward and confirmed to work well.

akien-mga:
Thanks for the amazing work on this BTW, it's a major improvement to our build pipeline! :heart:

Dockerfile.apple
```diff
@@ -0,0 +1,281 @@
...
+#   * Swift runtime + stdlib under
+#     .../Toolchains/XcodeDefault.xctoolchain/usr/lib/swift
+#   * Compiler-rt under
+#     .../Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/<ver>/lib/darwin/
+#     (libclang_rt.osx.a, libclang_rt.ios.a, libclang_rt.tvos.a,
+#     libclang_rt.xros.a, …)
+
+FROM base AS xcode-tree
+
+ENV XCODE_SDKV=26.6
```
akien-mga, line 87: Add `APPLE_SDKV=26.5` so we can use it in build scripts for iOS/visionOS.

Dockerfile.apple
```diff
@@ -0,0 +1,281 @@
...
+# darwin-tools-linux-llvm's ld64.lld instead (next step).
+
+FROM xcode-tree AS swift
+
+# Swift toolchain version: the swiftly identifier, which is also the installed
+# toolchain directory name that build-<platform>/build.sh composes into the
+# Swift compiler path. Override to build an experimental image, e.g.
+#   --build-arg swift_version=6.4.x-snapshot-2026-06-11
+# The default tracks the production release.
+ARG swift_version=6.3.2
```
akien-mga, line 112: Seems like Xcode 26.6 includes Swift 6.3.3, would be good to keep in sync.

Dockerfile.apple
```diff
@@ -0,0 +1,281 @@
...
+ENV SWIFT_VERSION=${swift_version}
```
akien-mga, line 113: Seems like we export it already so we can use it in the build scripts instead of hardcoding it.
We should also define `SWIFT_COMPILER` here for convenience so we don't have to repeat it in all build scripts.

Dockerfile.apple
```diff
@@ -0,0 +1,281 @@
...
+
+# swiftly 1.1.2 pinned: required to resolve newer toolchain branches such as
```
akien-mga, line 115: There seems to be a 1.1.3 now.
Pinning makes sense in general to keep builds reproducible.

Dockerfile.apple
```diff
@@ -0,0 +1,281 @@
...
+# darwin-tools-linux-llvm — Apple-platform-enabled ld64.lld
+# -----------------------------------------------------------------------------
+#
+# From xtool-org's prebuilt bundle. Provides:
+#   * ld64.lld  — LLD Mach-O driver with macOS/iOS/tvOS/visionOS enabled
+#   * dsymutil  — dSYM generation
+#   * libtool   — llvm-libtool-darwin (Darwin-format static archiver)
+#
+# Pinned to v1.0.1 (Dec 2024, LLD 20 from swiftlang/llvm-project).
+# Known limitation: LLD 20 cannot encode arm64e pauth relocations in
```
akien-mga, line 140:
```suggestion
# Known limitation: LLD 20 cannot encode arm64e path relocations in
```

Dockerfile.apple
```diff
@@ -0,0 +1,281 @@
...
+# -----------------------------------------------------------------------------
+# darwin-tools-linux-llvm — Apple-platform-enabled ld64.lld
+# -----------------------------------------------------------------------------
+#
+# From xtool-org's prebuilt bundle. Provides:
+#   * ld64.lld  — LLD Mach-O driver with macOS/iOS/tvOS/visionOS enabled
+#   * dsymutil  — dSYM generation
+#   * libtool   — llvm-libtool-darwin (Darwin-format static archiver)
+#
+# Pinned to v1.0.1 (Dec 2024, LLD 20 from swiftlang/llvm-project).
```
akien-mga, line 139: It's not super clear here, but I guess it's only pinned to that old version because it's the last one on https://github.com/xtool-org/darwin-tools-linux-llvm, not because newer version wouldn't work.

Dockerfile.apple
```diff
@@ -0,0 +1,281 @@
...
+#
+# platform/{ios,tvos,visionos}/detect.py returns True from can_build()
+# when the corresponding OSXCROSS_* env var is set. Value is arbitrary
+# (non-empty). OSXCROSS_APPLEEMBEDDED is referenced by the embedded
+# drivers for similar gates.
+
+ENV OSXCROSS_IOS=not_nothing
+ENV OSXCROSS_TVOS=not_nothing
+ENV OSXCROSS_VISIONOS=not_nothing
+ENV OSXCROSS_APPLEEMBEDDED=not_nothing
```
akien-mga, line 201: Since we don't use osxcross we should rework this. Using the new `APPLE_LLVM_CROSS` would be cleaner.

Dockerfile.apple
```diff
@@ -0,0 +1,281 @@
...
+# -----------------------------------------------------------------------------
+# ioscross-compatibility shims
+# -----------------------------------------------------------------------------
+#
+# Godot's platform/{ios,tvos,visionos}/detect.py composes compiler and
+# archiver paths as `${APPLE_TOOLCHAIN_PATH}/usr/bin/${apple_target_triple}<tool>`.
+# The existing build-{ios,tvos,visionos}/build.sh scripts set
+#   APPLE_TOOLCHAIN_PATH=/root/ioscross/<arch>
+#   apple_target_triple=<prefix>-apple-darwin11-
+# inherited from the old cctools-port wrapper layout.
```
akien-mga, line 212: Sounds like we could simplify things here and just not use the old `darwin11` triple in build scripts?

akien-mga (commented):
Found some time, so belated review of the new container.

Dockerfile.apple
akien-mga: Done in https://github.com/godotengine/build-containers/pull/167

Dockerfile.apple
akien-mga: Done in https://github.com/godotengine/build-containers/pull/167

Dockerfile.apple
akien-mga: Done in https://github.com/godotengine/build-containers/pull/167

Dockerfile.apple
akien-mga: Done in https://github.com/godotengine/build-containers/pull/167

Dockerfile.apple
akien-mga: Done in https://github.com/godotengine/build-containers/pull/167

Dockerfile.apple
akien-mga: Done in https://github.com/godotengine/build-containers/pull/167 and https://github.com/godotengine/godot/pull/121407

Dockerfile.apple
stuartcarnie: Yes, those are gone now from the upgrades scripts

PR #166, opened 2026-07-15
Update various toolchains for 4.8

akien-mga:
- Fedora 44
- .NET SDK 10.0
- Linux: SDKs godot-2026.05.x-1 with GCC 15.2.0, glibc 2.34, binutils 2.46.0
- Windows: MinGW 13.0.0 (unchanged) with GCC 16.1.1, binutils 2.46.0
- Android: NDK r29, platform 36
- Apple: Xcode 26.6 with Swift 6.3.3
- Web: Unchanged, Emscripten 4.0.20 pinned

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -1,18 +1,23 @@
 ARG img_version
 FROM godot-fedora:${img_version}
 
+# Keep NDK, JDK, build-tools and platforms in sync with Godot's
+# `platform/android/java/app/config.gradle`.
+
 ENV ANDROID_SDK_ROOT=/root/sdk
-ENV ANDROID_NDK_VERSION=28.1.13356709
+ENV ANDROID_NDK_VERSION=29.0.14206865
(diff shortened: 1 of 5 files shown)


Dockerfile.apple
```diff
@@ -109,7 +109,7 @@ FROM xcode-tree AS swift
 # Swift compiler path. Override to build an experimental image, e.g.
 #   --build-arg swift_version=6.4.x-snapshot-2026-06-11
 # The default tracks the production release.
-ARG swift_version=6.3.2
+ARG swift_version=6.3.3
```
akien-mga: FYI @stuartcarnie - seems like this is the version included in Xcode 26.6: https://en.wikipedia.org/wiki/Xcode#Xcode_26.0_(since_version_number_change)_2

Feel free to include this change in your own PR if you do further changes to match outstanding review feedback, as this PR handles mostly unrelated stuff and still needs testing.

akien-mga: Nevermind, I'm moving it to another PR.

akien-mga:
Tested via #167.

